PR #40, opened 2026-02-17
feat: add shell completions for tw CLI

gnapse:
Adds shell completion support for bash, zsh, and fish via a new `tw completion` command powered by `@pnpm/tabtab`, including install/uninstall flows and an internal completion server.

It introduces a Commander command-tree completion engine with option/value suggestions, wires `src/index.ts` for completion-server lazy loading, and adds completion value metadata for `search --type` and `thread reply --notify`.

This follows the same approach used in Todoist CLI: https://github.com/Doist/todoist-cli/pull/66. It also adds completion-focused tests and updates `README.md`, `src/lib/skills/content.ts`, and `AGENTS.md` so docs and skill content match the new command surface. Validation passed via format checks, type-check, and full tests (including lefthook pre-push tests).

## Demo

https://github.com/user-attachments/assets/7c181b23-0bc1-484f-89e6-84076288384d

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -4,59 +4,41 @@ import { type Command, program } from 'commander'
 import pkg from '../package.json' with { type: 'json' }
 import { startEarlySpinner, stopEarlySpinner } from './lib/spinner.js'
 
+const loadWorkspaceCommand = async () =>
+    (await import('./commands/workspace.js')).registerWorkspaceCommand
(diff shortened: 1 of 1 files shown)


gnapse:
> However, some of the commands fail for me (both on this feature branch and on main as well):

Nice catch. Fixed in the latest commit.

PR #121, opened 2026-04-01
fix: suppress credential manager warning on every command

scottlovegrove:
## Summary
- Removed the "system credential manager unavailable; using plaintext token from ..." warning that was printed to stderr on every command invocation when the secure store isn't available
- Users whose token is working fine via the config file fallback will no longer see alarming warnings on every single command
- The one-time warning at token save time (`login`/`token set`) is preserved so users are still informed when their token falls back to plaintext storage

## Test plan
- [x] Verified all 321 tests pass
- [x] Updated the test that previously asserted the warning was printed to instead assert `errorSpy.not.toHaveBeenCalled()`
- [ ] Manual: run `tw` commands in an environment without a system credential manager and confirm no warning on each invocation
- [ ] Manual: run `tw auth token <token>` in the same environment and confirm the one-time fallback warning still appears

### DIFF
--- a/src/__tests__/lib/auth.test.ts
+++ b/src/__tests__/lib/auth.test.ts
@@ -198,7 +198,7 @@ describe('auth token storage', () => {
         })
     })
 
-    it('reads from plaintext config with a warning when the secure store is unavailable', async () => {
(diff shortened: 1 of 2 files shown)


henningmu:
Thank you ✅

PR #126, opened 2026-04-02
feat: allow switching between stable and pre-release update channels

scottlovegrove:
## Summary
- Converts `src/commands/update.ts` into a `src/commands/update/` module with `action.ts`, `channel.ts`, `switch.ts`, and `index.ts`
- Adds `tw update switch --stable` / `tw update switch --pre-release` commands to opt into pre-release updates from the CLI
- Adds `tw update channel` command to show the current update channel
- The update command now reads `update_channel` from config and fetches from the appropriate npm dist-tag (`latest` vs `next`), with semver-aware comparison that handles downgrades gracefully

## Test plan
- [x] `npm run type-check` passes
- [x] All 334 tests pass (`npm test`)
- [x] `npm run lint:check` passes (oxlint + oxfmt)
- [x] `npm run check:skill-sync` confirms SKILL.md is in sync
- [ ] Verify `tw update switch --pre-release` sets config and shows warning
- [ ] Verify `tw update switch --stable` resets config
- [ ] Verify `tw update channel` shows the correct channel
- [ ] Verify `tw update` fetches from `/next` when on pre-release channel
- [ ] Verify `tw update --check` shows channel info

### DIFF
--- a/src/commands/update/action.ts
+++ b/src/commands/update/action.ts
@@ -27,7 +27,8 @@ function isNewer(current: string, candidate: string): boolean {
     }
     if (!a.prerelease && b.prerelease) return false
     if (a.prerelease && !b.prerelease) return true
-    if (a.prerelease && b.prerelease) return b.prerelease > a.prerelease
+    if (a.prerelease && b.prerelease)
+        return b.prerelease.localeCompare(a.prerelease, undefined, { numeric: true }) > 0
     return false
 }
 
@@ -73,7 +74,7 @@ interface UpdateOptions {
 export async function updateAction(options: UpdateOptions): Promise<void> {
     const currentVersion = pkg.version
     const config = await getConfig()
-    const channel = config.update_channel ?? 'stable'
+    const channel = config.updateChannel ?? 'stable'
     const tag = channel === 'pre-release' ? 'next' : 'latest'
     const label = channel === 'pre-release' ? ` ${chalk.magenta('(pre-release)')}` : ''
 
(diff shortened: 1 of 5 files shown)


src/commands/update/action.ts
```diff
@@ -1,14 +1,39 @@
...
+
+function isNewer(current: string, candidate: string): boolean {
+    const a = parseVersion(current)
+    const b = parseVersion(candidate)
+    for (const key of ['major', 'minor', 'patch'] as const) {
+        if (b[key] !== a[key]) return b[key] > a[key]
+    }
+    if (!a.prerelease && b.prerelease) return false
+    if (a.prerelease && !b.prerelease) return true
+    if (a.prerelease && b.prerelease) return b.prerelease > a.prerelease
```
doistbot: ### [P1] Incorrect Prerelease Version Comparison

Using a standard string comparison for prerelease tags causes logical bugs with numeric suffixes. For example, `"rc.10" < "rc.2"` evaluates to `true` in JavaScript, so the CLI will incorrectly treat `rc.10` as an older version and present the downgrade warning message instead of an update available message.

Use `localeCompare` with `numeric: true` to ensure the numeric parts of the prerelease tag are evaluated correctly.

```suggestion
    if (a.prerelease && b.prerelease) return b.prerelease.localeCompare(a.prerelease, undefined, { numeric: true, sensitivity: 'base' }) > 0
```

src/lib/config.ts
```diff
@@ -15,6 +16,7 @@ export interface Config {
     // Auth metadata persisted alongside the token to track OAuth scope.
     authMode?: AuthMode
     authScope?: string
+    update_channel?: UpdateChannel
```
doistbot: ### [P3] Inconsistent Naming Convention

The property `update_channel` uses `snake_case`, which is inconsistent with the `camelCase` naming convention used by the other properties in the `Config` interface (e.g., `currentWorkspace`, `authMode`, `authScope`). Consider renaming it to `updateChannel` for consistency.

```suggestion
    updateChannel?: UpdateChannel
```

src/commands/update/action.ts
```diff
@@ -63,11 +94,23 @@ export async function updateAction(options: UpdateOptions): Promise<void> {
...
 
+    if (!newer) {
+        console.log(
+            chalk.yellow('Note:'),
+            `v${latestVersion}${channel === 'pre-release' ? ' (pre-release)' : ''} is older than your current v${currentVersion}`,
+        )
+    }
+
     if (options.check) {
+        const channelLabel =
```
doistbot: [P2] Channel info is still skipped on the up-to-date `--check` path. This block only runs after the `Update available` branch, so when `currentVersion === latestVersion` the function returns earlier and `tw update --check` omits the configured channel entirely. If `--check` is meant to always show channel info now, this output needs to move before the equality early return or be duplicated in that branch.

PR #159, opened 2026-04-08
fix: surface batch API errors in commands

### LINKED ISSUE #158: Some commands don’t handle errors and crash with misleading internal errors
Some `tw` commands appear to assume API/batch responses always contain successful `.data` payloads.

Because of that, when an API call fails, the CLI can crash with a misleading internal error instead of surfacing the real validation or request error.

## Example

```bash
tw inbox --unread --limit 1000
```

The API correctly rejects this because `limit` must be `<= 500`, but instead of showing that cleanly, the CLI crashes with:

```txt
TypeError: threads.data.map is not a function
```

## Why this seems more general

In some commands, the code immediately accesses things like:

```js
threads.data.map(...)
response.data.id
```

without first checking that the API call succeeded.

So this looks broader than an inbox-specific bug.

## Expected

If an API call fails, the CLI should surface the actual API error cleanly instead of crashing with a misleading internal `TypeError`.

## Versions

- `@doist/twist-cli`: `2.26.0`
- `@doist/twist-sdk`: `2.3.0`

goncalossilva:
Closes #158. The issue includes a repro.


src/commands/thread/view.ts
```diff
@@ -42,9 +42,7 @@ async function viewSingleComment(
     )
 
     const channel = assertBatchData(channelResponse, 'channel')
-    const userMap = new Map(
-        userResponses.filter((r) => r.data != null).map((r) => [r.data.id, r.data.name]),
-    )
+    const userMap = buildBatchNameMap(userIds, userResponses, 'user')
```
doistbot: [P1] This changes the thread-view user lookup path from the previous `filter((r) => r.data != null)` behavior to a hard failure. The surrounding rendering code still expects missing map entries and falls back to `user:${id}`, so a `null` user batch result now aborts `tw thread view` instead of rendering the thread. Please preserve the null-skipping behavior here (and in the similar call below) while still surfacing real error payloads.

doistbot (commented):
This PR updates the CLI commands to properly surface batch API errors, improving observability and error awareness. However, the updated user lookup logic in the thread view inadvertently causes a hard failure for missing user entries instead of preserving the existing fallback behavior that gracefully skips null results.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F159&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A8ee6b950-337d-11f1-94e0-00263d6d6bcb%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A8ee6b950-337d-11f1-94e0-00263d6d6bcb&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove (approved):

src/commands/thread/view.ts
scottlovegrove: Addressed in 606b6b8. Added `buildOptionalBatchNameMap` in `src/lib/api.ts` which skips entries with `data: null` and a success code (preserving the prior null-skipping behavior) while still throwing on real API errors via `assertBatchData`. Switched the user lookups in `thread/view.ts`, `conversation/view.ts`, `conversation/unread.ts`, and `conversation/helpers.ts` to the tolerant variant; channels stay strict via `buildBatchNameMap`. Added a regression test in `thread.test.ts` covering the `{ code: 200, data: null }` user case, and updated AGENTS.md to document when to use each helper.

PR #197, opened 2026-04-28
feat: add mentions command and search pagination

gnapse:
This adds a dedicated `tw mentions` command for fetching content that mentions the current user, with the same filters and output modes as search plus an `--all` option to walk every results page.

Mentions were previously only reachable through `tw search --mention-me`, which still required a query string and left full backlog traversal to manual cursor handling.

The change also refactors shared search plumbing so `tw search` reuses the same resolver and output path, while the CLI docs and generated Twist skill content now advertise both `tw mentions` and `tw search --all`.

Validation: `npm test`, `npx vitest run src/commands/search.test.ts src/commands/mentions.test.ts`, `npm run type-check`, `npm run lint:check`, and `npm run check:skill-sync`.

### DIFF
--- a/src/commands/mentions.ts
+++ b/src/commands/mentions.ts
@@ -1,21 +1,21 @@
 import { Command } from 'commander'
 import {
     addSharedSearchOptions,
-    printSearchCommandResults,
-    runSearchCommand,
-    type SearchCommandOptions,
-} from '../lib/search-command.js'
+    printSearchResults,
+    runSearch,
+    type SharedSearchOptions,
+} from '../lib/search-helpers.js'
 
 async function mentions(
     workspaceRef: string | undefined,
-    options: SearchCommandOptions,
+    options: SharedSearchOptions,
 ): Promise<void> {
-    const { workspaceId, response } = await runSearchCommand(workspaceRef, {
+    const { workspaceId, response } = await runSearch(workspaceRef, {
         ...options,
         mentionSelf: true,
     })
 
-    printSearchCommandResults(workspaceId, response, options)
+    printSearchResults(workspaceId, response, options)
 }
 
 export function registerMentionsCommand(program: Command): void {
--- a/src/commands/mentions.ts
+++ b/src/commands/mentions.ts
@@ -1,6 +1,6 @@
-import { Command, Option } from 'commander'
-import { withCaseInsensitiveChoices } from '../lib/completion.js'
+import { Command } from 'commander'
 import {
+    addSharedSearchOptions,
     printSearchCommandResults,
     runSearchCommand,
     type SearchCommandOptions,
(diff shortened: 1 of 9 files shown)


src/lib/search-command.ts
```diff
@@ -0,0 +1,267 @@
...
+}
+
+type SearchOutputOptions = Pick<SearchCommandOptions, 'all' | 'json' | 'ndjson' | 'full'>
+
+export function printSearchCommandResults(
+    workspaceId: number,
+    response: ExtendedSearchResponse,
+    options: SearchOutputOptions,
+): void {
+    if (response.items.length === 0) {
```
doistbot: [P1] Empty result sets still bypass the machine-readable branches. Because this early return runs before the `json`/`ndjson` handling, `tw mentions --json` / `--ndjson` with no matches emits `No results found.` (or the public-results notice) instead of parseable output, which breaks scripting for the new command. Handle the structured output modes first and emit an empty `{ results: [], nextCursor: ... }` / `_meta` payload when there are no items.

src/commands/mentions.ts
```diff
@@ -0,0 +1,53 @@
...
+    })
+
+    printSearchCommandResults(workspaceId, response, options)
+}
+
+export function registerMentionsCommand(program: Command): void {
+    program
+        .command('mentions [workspace-ref]')
+        .description('Show content mentioning the current user')
+        .option('--workspace <ref>', 'Workspace ID or name')
```
doistbot: [P2] The search filter/output flags are now duplicated across `search.ts` and `mentions.ts`. Since the runtime logic was already centralized in `search-command.ts`, keeping two separate option lists means future flag/help changes have to be mirrored manually in both commands (and they are already starting to diverge in this diff). Consider extracting a shared helper that adds the common search options/help text so the two commands stay in sync.

doistbot (commented):
This introduces the `tw mentions` command alongside new `--all` pagination for search, cleanly centralizing the underlying search resolution logic. Reusing the output path effectively enhances the maintainability of the CLI's search capabilities. A few adjustments are needed to ensure empty result sets still return valid structured objects in JSON and NDJSON modes, and to deduplicate the shared search CLI flags to prevent future drift between the commands.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F197&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A553f5500-434e-11f1-85ec-72590f2d68a0%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A553f5500-434e-11f1-85ec-72590f2d68a0&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/search-command.ts
```diff
@@ -0,0 +1,308 @@
...
+    options: SearchRequestOptions,
+): Promise<SearchCommandResult> {
+    const workspaceId = await resolveWorkspaceId(workspaceRef, options.workspace)
+    const params = await buildSearchParams(workspaceId, options)
+    const response = await fetchSearchPages(params, options.all)
+    return {
+        workspaceId,
+        response: await filterVisibleSearchResults(workspaceId, response),
+    }
+}
```
scottlovegrove: Not a huge fan of this being exported and imported into a bunch of other commands. Feels like it should be a helper function somewhere that all commands can use, rather than importing from search-commands. 

Not a hill I will die on though.

scottlovegrove (approved):

src/lib/search-command.ts
gnapse: I agree with this feedback.

PR #202, opened 2026-05-06
feat(groups): add membership and CRUD subcommands

lmjabreu:
## Summary

Wraps the SDK group endpoints as new subcommands under `tw groups`. Workspace admins can now manage roster cleanup from the CLI without dropping into the Twist web UI.

| Command | Description |
|---|---|
| `tw groups` / `tw groups list [workspace-ref]` | List groups (existing behavior, default subcommand) |
| `tw groups view <group-ref>` | Show a group with its members (names + emails) |
| `tw groups create <name> [--users <refs>]` | Create a group, optionally seeding members |
| `tw groups rename <group-ref> <new-name>` | Rename a group |
| `tw groups delete <group-ref> [--yes]` | Delete a group (confirmation prompt unless `--yes`) |
| `tw groups add-user <group-ref> <user-refs...>` | Add users to a group |
| `tw groups remove-user <group-ref> <user-refs...>` | Remove users from a group |

`add-user` / `remove-user` accept user references space-separated, comma-separated within a single arg, or a mix. Both use the SDK's bulk `addUsers` / `removeUsers` endpoints, so each command is a single API call regardless of how many users you pass.

### Group ref resolution

New `resolveGroupRef(ref, workspaceId)` mirrors `resolveUserRefs`:
- `id:N` or bare numeric → ID lookup
- name → exact (case-insensitive) preferred, then unique substring match
- ambiguous name → `AMBIGUOUS_GROUP` error listing matches with IDs (use ID to disambiguate)
- not found → `GROUP_NOT_FOUND` with a hint to run `tw groups`

### OAuth scopes — three-tier model

Twist uses a **three-tier scope model** for groups: `:read` / `:write` / `:remove`. This is different from most other Twist resources which only use `:read` / `:write`.

- `groups:read` — list and view groups (already present)
- `groups:write` — create groups, rename groups, add users to groups
- `groups:remove` — **delete groups, remove users from groups**

Both `groups:write` and `groups:remove` are now added to `READ_WRITE_SCOPES`. The `:remove` scope was discovered during live testing — the `remove-user` endpoint returned 403 with only `groups:write`. **Users with existing read-write tokens need to re-authenticate (`tw auth login`) to pick up the new scopes.**

### Out of scope (intentionally)

- Bulk imports from CSV
- "Sync from another group"
- Interactive multi-select pickers
- Group `description` / `color` fields (the SDK supports them, but the requested scope was just membership and CRUD; happy to add in a follow-up)

### CHANGELOG note

This repo uses semantic-release to generate `CHANGELOG.md` from conventional commits, so I didn't add a manual entry — the `feat(groups):` commit will produce one on the next release.

## Test plan

- [x] `npm run lint` clean
- [x] `npm test` — 569 passed (added 21 new tests covering ref resolution, ambiguity, command wiring, and the dedupe logic in `add-user` / `remove-user`)
- [x] `npm run build` clean
- [x] Live tested `tw groups remove-user "Component Library" "Stephen Barkan" "Stijn" "Julia Hilt"` — successfully removed 3 members (13 → 10)


doistbot (commented):
This is a fantastic addition that brings much-needed group CRUD and membership management to the CLI, streamlining roster maintenance for workspace admins. The implementation effectively leverages the SDK's bulk endpoints to keep operations efficient. A few adjustments are needed to fully align with our established patterns, particularly around updating the AI skill content and error codes, optimizing redundant and sequential API calls, refining the Commander arguments to preserve structured errors, and consolidating shared logic and test coverage.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F202&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A66e3acf0-4959-11f1-98e6-436d1ea34580%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A66e3acf0-4959-11f1-98e6-436d1ea34580&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/groups/index.ts
```diff
@@ -0,0 +1,126 @@
+import { Command } from 'commander'
+import { createGroupCommand } from './create.js'
+import { deleteGroupCommand } from './delete.js'
+import { listGroups } from './list.js'
+import { addUsersCommand, removeUsersCommand } from './members.js'
+import { renameGroup } from './rename.js'
+import { viewGroup } from './view.js'
+
+export function registerGroupsCommand(program: Command): void {
```
doistbot, line 9: [P1] `AGENTS.md` explicitly requires that `src/lib/skills/content.ts` be updated whenever commands or subcommands are added or modified. The new `groups` subcommands (`view`, `create`, `rename`, `delete`, `add-user`, `remove-user`) must be documented in `SKILL_CONTENT` so that installed AI agents are aware of them.

src/commands/groups/create.ts
```diff
@@ -0,0 +1,41 @@
...
+import { resolveUserRefs, resolveWorkspaceRef } from '../../lib/refs.js'
+
+type CreateGroupOptions = MutationOptions & {
+    workspace?: string
+    users?: string
+}
+
+export async function createGroupCommand(name: string, options: CreateGroupOptions): Promise<void> {
+    if (!name || name.trim() === '') {
+        throw new CliError('INVALID_NAME', 'Group name cannot be empty.')
```
doistbot, line 14: [P2] `AGENTS.md` requires that new error codes be explicitly added to the `ErrorCode` type in `src/lib/errors.ts`. `INVALID_NAME` (along with `MISSING_USERS`, `GROUP_NOT_FOUND`, and `AMBIGUOUS_GROUP` introduced in this PR) needs to be registered there.

src/commands/groups/index.ts
```diff
@@ -0,0 +1,126 @@
...
+            'after',
+            `
+Examples:
+  tw groups delete 12345 --yes
+  tw groups delete "Frontend" --dry-run`,
+        )
+        .action(deleteGroupCommand)
+
+    groups
+        .command('add-user <group-ref> <user-refs...>')
```
doistbot: [P2] Declaring this as `<user-refs...>` means Commander throws before `joinUserRefs()` can run when no refs are provided. That makes your `MISSING_USERS` `CliError` unreachable here (and for `remove-user` below), so `--json` mode falls back to a generic `INTERNAL_ERROR` instead of the structured validation error you added. Make the variadic arg optional and keep the validation in the handler if you want that error path to work.

src/commands/groups/members.ts
```diff
@@ -0,0 +1,132 @@
...
+}
+
+export async function addUsersCommand(
+    groupRef: string,
+    userRefs: string[],
+    options: MutationOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const group = await resolveGroupRef(groupRef, workspaceId)
+    const userIds = await resolveUserRefs(joinUserRefs(userRefs), workspaceId)
```
doistbot: [P2] `resolveUserRefs()` preserves duplicates, so inputs like `id:3,id:3` or `alice alice@doist.com` can produce duplicate IDs here. `newIds`/`presentIds` will then overcount the change (`Added 2 user(s)` / `Removed 2 user(s)`) and may send duplicate IDs to the bulk SDK call. Deduplicate the resolved IDs before filtering and invoking the API.

src/lib/refs.ts
```diff
@@ -347,6 +347,45 @@ export function parseUserIdRefs(refs: string): number[] {
...
+        const group = groups.find((g) => g.id === parsed.id)
+        if (!group) {
+            throw new CliError('GROUP_NOT_FOUND', `Group with ID ${parsed.id} not found`, [
+                'Run: tw groups to list available groups',
+            ])
+        }
+        return group
+    }
+
+    if (parsed.type === 'name') {
```
doistbot, line 386: [P2] `resolveGroupRef` reimplements the same exact-match / unique-substring / ambiguous-match flow that already exists in `resolveWorkspaceRef` and `resolveChannelRef` in this file. Pulling that name-matching/error-formatting logic into a shared helper would keep entity ref resolution consistent instead of adding a third copy to maintain.

src/commands/groups/members.ts
```diff
@@ -0,0 +1,132 @@
...
+        console.log(
+            `Added ${newIds.length} user(s) to "${updated.name}" (now ${updated.userIds.length} members).`,
+        )
+    }
+    if (alreadyIds.length > 0) {
+        console.log(`Already members: ${alreadyIds.join(', ')}`)
+    }
+}
+
+export async function removeUsersCommand(
```
doistbot: [P2] `removeUsersCommand` is nearly the same control flow as `addUsersCommand` above: resolve workspace/group/users, partition IDs against current members, handle dry-run, call the SDK, refetch, then format JSON/text output. Extracting the shared “mutate group membership” path into a private helper would reduce duplication and keep the add/remove subcommands from drifting as their output/options evolve.

src/commands/groups/view.ts
```diff
@@ -0,0 +1,48 @@
...
+            id,
+            name: user?.name ?? null,
+            email: user?.email ?? null,
+        }
+    })
+
+    const enriched = { ...group, members }
+
+    if (options.json) {
+        console.log(formatJson(enriched))
```
doistbot: [P2] This bypasses the CLI's normal machine-readable contract: because `view` has no `--full` flag and this call omits an entity type, `tw groups view --json/--ndjson` always emits the raw SDK shape (plus `members`), while `tw groups --json` returns filtered group fields. That inconsistent schema makes automation brittle. Please add `--full` and route group-view output through an explicit default shape instead of always serializing the entire object.

src/commands/groups/view.ts
```diff
@@ -0,0 +1,48 @@
...
+import type { ViewOptions } from '../../lib/options.js'
+import { colors, formatJson, formatNdjson, pluralize } from '../../lib/output.js'
+import { resolveGroupRef } from '../../lib/refs.js'
+
+type GroupViewOptions = ViewOptions
+
+export async function viewGroup(ref: string, options: GroupViewOptions): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const summary = await resolveGroupRef(ref, workspaceId)
+    const group = await getGroup(summary.id)
```
doistbot: [P2] `resolveGroupRef` already returns a complete `Group` object. The subsequent `getGroup(summary.id)` call is a redundant network request. Additionally, `getWorkspaceUsers` is an independent read and can be fetched concurrently:

```typescript
    const [group, users] = await Promise.all([
        resolveGroupRef(ref, workspaceId),
        getWorkspaceUsers(workspaceId)
    ])
```

src/commands/groups/members.ts
```diff
@@ -0,0 +1,132 @@
...
+    return userRefs.join(',')
+}
+
+export async function addUsersCommand(
+    groupRef: string,
+    userRefs: string[],
+    options: MutationOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const group = await resolveGroupRef(groupRef, workspaceId)
```
doistbot: [P2] `resolveGroupRef` and `resolveUserRefs` perform independent API requests (`getWorkspaceGroups` and `getWorkspaceUsers`). Executing them sequentially adds unnecessary latency. They should be run concurrently using `Promise.all`:

```typescript
    const [group, userIds] = await Promise.all([
        resolveGroupRef(groupRef, workspaceId),
        resolveUserRefs(joinUserRefs(userRefs), workspaceId),
    ])
```
(This also applies to `removeUsersCommand` on lines 84-85).

src/commands/groups/members.ts
```diff
@@ -0,0 +1,132 @@
...
+            'Already members': alreadyIds.length > 0 ? alreadyIds.join(', ') : undefined,
+        })
+        return
+    }
+
+    if (newIds.length > 0) {
+        await addUsersToGroup(group.id, newIds)
+    }
+
+    const updated = await getGroup(group.id)
```
doistbot: [P2] Fetching the updated group via `getGroup(group.id)` adds an avoidable API call. This fetch is only necessary if `options.full` is provided to return the complete object. For standard text output and minimal JSON, the new state is deterministic and can be computed locally (e.g. `group.userIds.length + newIds.length`), using `group.name` for the output. This optimization applies to both `addUsersCommand` and `removeUsersCommand`.

src/lib/refs.ts
```diff
@@ -347,6 +347,45 @@ export function parseUserIdRefs(refs: string): number[] {
     })
 }
 
+export async function resolveGroupRef(ref: string, workspaceId: number): Promise<Group> {
+    const { getWorkspaceGroups } = await import('./api.js')
+    const parsed = parseRef(ref)
+
+    if (parsed.type === 'id') {
+        const groups = await getWorkspaceGroups(workspaceId)
```
doistbot: [P2] The numeric/id branch now resolves a group by loading the entire workspace group list and scanning it. That makes every `tw groups view|rename|delete|add-user|remove-user id:...` call pay for `getGroups` even though `getGroup(id)` already exists. Prefer the narrower lookup here (and verify `workspaceId` on the returned group if you still need workspace scoping) to avoid O(all groups) reads on the ID path.

src/commands/groups/groups.test.ts
```diff
@@ -0,0 +1,365 @@
...
+            userIds: undefined,
+        })
+        expect(consoleSpy.mock.calls[0][0]).toContain('Design')
+        consoleSpy.mockRestore()
+    })
+
+    it('resolves --users and passes ids to createGroup', async () => {
+        refsMocks.resolveUserRefs.mockResolvedValue([10, 20])
+        const program = createProgram()
+        vi.spyOn(console, 'log').mockImplementation(() => {})
```
doistbot, line 241: [P3] `vi.spyOn(console, 'log')` is overridden here (and in several subsequent tests) without assigning it or restoring it. Because `beforeEach` only calls `vi.clearAllMocks()` (which resets call counts but not implementations), this will permanently silence `console.log` for any remaining tests in the file. Consider capturing the spy and calling `.mockRestore()` like earlier tests do, or use `vi.restoreAllMocks()` in `beforeEach`.

src/commands/groups/groups.test.ts
```diff
@@ -0,0 +1,365 @@
...
+        description: 'Cross-team',
+        workspaceId: 1,
+        userIds: [1, 4],
+        version: 1,
+    },
+]
+
+const frontend = sampleGroups[0]
+
+describe('tw groups list (default)', () => {
```
doistbot: [P2] This refactor changes `groups` from a flat command to a default subcommand, but the moved suite no longer locks in the previously-covered list behavior around `[workspace-ref]`, `--search`, empty results, `--ndjson`, and `--full`. The new block only exercises bare `tw groups`, explicit `list`, and basic JSON, so an existing-list regression could now ship unnoticed.

src/commands/groups/groups.test.ts
```diff
@@ -0,0 +1,365 @@
...
+
+    it('errors in --json mode without --yes', async () => {
+        const program = createProgram()
+        await expect(
+            program.parseAsync(['node', 'tw', 'groups', 'delete', 'Frontend', '--json']),
+        ).rejects.toMatchObject({ code: 'MISSING_YES_FLAG' })
+    })
+})
+
+describe('tw groups add-user', () => {
```
doistbot, line 318: [P2] The new membership coverage only checks users who are already in/out of the group; it never passes the same ref twice in one invocation (for example `add-user Frontend id:3 id:3` or `remove-user ... id:2,id:2`). Because these commands forward a bulk `userIds` array to the SDK, please add an explicit repeated-ref case so the suite catches duplicate IDs being sent unchanged.

src/commands/groups/index.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/create.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/index.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/members.ts
lmjabreu: Addressed in 24c7da0.

src/lib/refs.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/members.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/view.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/view.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/members.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/members.ts
lmjabreu: Addressed in 24c7da0.

src/lib/refs.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/groups.test.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/groups.test.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/groups.test.ts
lmjabreu: Addressed in 24c7da0.

src/commands/groups/members.ts
```diff
@@ -0,0 +1,124 @@
...
+    // Compute the new member count locally instead of refetching
+    const newMemberCount =
+        action === 'add'
+            ? group.userIds.length + actionableIds.length
+            : group.userIds.length - actionableIds.length
+
+    if (options.json) {
+        if (options.full) {
+            // For --full we need the actual updated group from the API
+            const { getGroup } = await import('../../lib/api.js')
```
scottlovegrove: Why are using an inline import here? Why not just import it at the top?

src/commands/groups/view.ts
```diff
@@ -0,0 +1,58 @@
...
+    const userMap = new Map(users.map((u) => [u.id, u]))
+
+    const members = group.userIds.map((id) => {
+        const user = userMap.get(id)
+        return {
+            id,
+            name: user?.name ?? null,
+            email: user?.email ?? null,
+        }
+    })
```
scottlovegrove: Why are we fetching _all_ workspace users when we only want the ones in the group. We already have the list of user ids, so we should do a batch call to fetch the users, this way we only pull down the actual users we are interested in. 

Imagine pulling down all 100+ doisters to see a group with 2 people in it.

src/lib/refs.ts
```diff
@@ -347,6 +362,43 @@ export function parseUserIdRefs(refs: string): number[] {
     })
 }
 
+export async function resolveGroupRef(ref: string, workspaceId: number): Promise<Group> {
+    const { getGroup, getWorkspaceGroups } = await import('./api.js')
```
scottlovegrove: Again, no need to inline import, just import at the top.

scottlovegrove (changes requested):

src/commands/groups/members.ts
lmjabreu: Moved to top-level import.

src/commands/groups/view.ts
lmjabreu: Good call — switched to batch `getUserById` calls so we only fetch the users in the group, not the entire workspace.

src/lib/refs.ts
scottlovegrove: This hasn't been addressed yet @lmjabreu

lmjabreu: Moved to top-level import — no circular dependency here since api.ts doesn't import from refs.ts.

scottlovegrove (approved):

PR #205, opened 2026-05-08
fix: emit empty JSON/NDJSON when inbox/unread results empty

scottlovegrove:
## Summary
- `tw inbox --json` (and `--ndjson`) printed `No threads in inbox.` instead of `[]` when empty, breaking pipelines that parse the output as JSON. Same bug existed in `tw conversation unread --json`.
- Gate the empty-state human messages behind the non-machine-output branch so `--json` emits `[]` and `--ndjson` emits nothing.
- Added regression tests covering both commands for `--json`, `--ndjson`, and the human/text fallback. Also covers the inbox `--channel <filter>` no-match path.

Audited every other `--json`-supporting command (workspace, users, groups list, channel list, channel threads, conversation view/with/helpers, thread view, search/mentions) — they already gate JSON/NDJSON before any human empty message.

## Test plan
- [x] `npm run type-check`
- [x] `npx vitest run src/commands/inbox.test.ts src/commands/conversation/conversation.test.ts`
- [ ] Manual: `tw inbox --json` against an empty inbox → `[]`
- [ ] Manual: `tw inbox --ndjson` against an empty inbox → empty
- [ ] Manual: `tw conversation unread --json` with no unread → `[]`
- [ ] Manual: `tw inbox --json | jq '. | length'` → `0`

### DIFF
--- a/src/commands/conversation/conversation.test.ts
+++ b/src/commands/conversation/conversation.test.ts
@@ -252,9 +252,9 @@ describe('conversation unread empty output', () => {
         const program = createProgram()
         await program.parseAsync(['node', 'tw', 'conversation', 'unread', '--ndjson'])
 
-        const output = logSpy.mock.calls.map((c: unknown[]) => c[0]).join('\n')
-        expect(output.trim()).toBe('')
-        expect(output).not.toContain('No unread conversations')
+        // Must not call console.log at all — `console.log('')` still emits a
+        // stray newline and would break strict NDJSON consumers.
+        expect(logSpy).not.toHaveBeenCalled()
     })
 
     it('still prints human message when --json/--ndjson not set', async () => {
--- a/src/commands/conversation/conversation.test.ts
+++ b/src/commands/conversation/conversation.test.ts
@@ -1,6 +1,7 @@
 import { Command } from 'commander'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { CliError } from '../../lib/errors.js'
+import { describeEmptyMachineOutput } from '../../test-helpers/empty-output.js'
 
 const apiMocks = vi.hoisted(() => ({
     getTwistClient: vi.fn(),
@@ -228,42 +229,18 @@ describe('conversation unread --workspace conflict', () => {
     })
 })
 
-describe('conversation unread empty output', () => {
-    let logSpy: ReturnType<typeof vi.spyOn>
-
-    beforeEach(() => {
+describeEmptyMachineOutput('conversation unread empty output', {
+    setup: () => {
         vi.clearAllMocks()
         const client = createClient({})
         client.conversations.getUnread = vi.fn().mockResolvedValue([])
         apiMocks.getTwistClient.mockResolvedValue(client)
-        logSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
-    })
-
-    it('outputs [] for --json when there are no unread conversations', async () => {
+    },
+    run: async (extraArgs) => {
         const program = createProgram()
-        await program.parseAsync(['node', 'tw', 'conversation', 'unread', '--json'])
-
-        const output = logSpy.mock.calls.map((c: unknown[]) => c[0]).join('\n')
-        expect(output.trim()).toBe('[]')
-        expect(output).not.toContain('No unread conversations')
-    })
-
-    it('outputs nothing for --ndjson when there are no unread conversations', async () => {
-        const program = createProgram()
-        await program.parseAsync(['node', 'tw', 'conversation', 'unread', '--ndjson'])
-
-        // Must not call console.log at all — `console.log('')` still emits a
-        // stray newline and would break strict NDJSON consumers.
-        expect(logSpy).not.toHaveBeenCalled()
-    })
-
-    it('still prints human message when --json/--ndjson not set', async () => {
-        const program = createProgram()
-        await program.parseAsync(['node', 'tw', 'conversation', 'unread'])
-
-        const output = logSpy.mock.calls.map((c: unknown[]) => c[0]).join('\n')
-        expect(output).toContain('No unread conversations.')
-    })
+        await program.parseAsync(['node', 'tw', 'conversation', 'unread', ...extraArgs])
+    },
+    humanMessage: 'No unread conversations.',
 })
 
 describe('conversation with', () => {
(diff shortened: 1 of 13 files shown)


src/commands/conversation/unread.ts
```diff
@@ -31,6 +31,14 @@ export async function showUnread(
     const unreadConversations = await client.conversations.getUnread(workspaceId)
 
     if (unreadConversations.length === 0) {
+        if (options.json) {
+            console.log(formatJson([], 'conversation', options.full))
+            return
+        }
+        if (options.ndjson) {
+            console.log(formatNdjson([], 'conversation', options.full))
```
doistbot: [P1] `console.log(formatNdjson([], ...))` still writes a bare newline, so `tw conversation unread --ndjson` is not actually empty output. To match the new contract and avoid breaking strict NDJSON consumers, return without logging when the array is empty (or write to stdout only if the formatted string is non-empty).

src/commands/inbox.ts
```diff
@@ -64,6 +64,14 @@ async function showInbox(workspaceRef: string | undefined, options: InboxOptions
     }
 
     if (inboxThreads.length === 0) {
+        if (options.json) {
```
doistbot: [P2] The JSON/NDJSON empty state logic is duplicated three times in this file for different filter stages. Consider extracting a local helper closure to reduce boilerplate:

```typescript
const exitEmpty = (msg: string) => {
    if (options.json) console.log(formatJson([], 'thread', options.full))
    else if (options.ndjson) console.log(formatNdjson([], 'thread', options.full))
    else console.log(msg)
}
```

Then each check becomes a clean one-liner: `if (inboxThreads.length === 0) return exitEmpty('No threads in inbox.')`

src/commands/conversation/conversation.test.ts
```diff
@@ -228,6 +228,44 @@ describe('conversation unread --workspace conflict', () => {
...
+    it('outputs [] for --json when there are no unread conversations', async () => {
+        const program = createProgram()
+        await program.parseAsync(['node', 'tw', 'conversation', 'unread', '--json'])
+
+        const output = logSpy.mock.calls.map((c: unknown[]) => c[0]).join('\n')
+        expect(output.trim()).toBe('[]')
+        expect(output).not.toContain('No unread conversations')
+    })
+
+    it('outputs nothing for --ndjson when there are no unread conversations', async () => {
```
doistbot: [P1] This test doesn't actually verify "no output" at the CLI boundary. `console.log('')` still writes a newline to stdout, but `logSpy.mock.calls.map((c) => c[0]).join('\n')` reduces that to `''`, so the test passes even if the command emits a blank line. Please assert on `process.stdout.write` (or assert `console.log` is not called) so the regression catches a stray newline.

doistbot (commented):
This PR addresses the empty state behavior for the inbox and conversation unread commands, ensuring they correctly emit empty arrays for JSON and suppress output for NDJSON. This is a great fix that will effectively prevent human-readable fallback messages from breaking machine-readable pipelines. There are just a few areas to refine regarding stray newlines in the NDJSON output, strengthening the test assertions to accurately catch those newlines, and reducing boilerplate by extracting the duplicated empty-state logic into a helper function.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F205&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Adda4e740-4af9-11f1-96be-28de1dd3dfd3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Adda4e740-4af9-11f1-96be-28de1dd3dfd3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/conversation/unread.ts
scottlovegrove: Fixed in 61ebe83. The new `printEmpty` helper does not call `console.log` at all in the `--ndjson` empty case, so no stray newline is emitted. `tw conversation unread --ndjson` now produces zero bytes when there are no unread conversations.

src/commands/inbox.ts
scottlovegrove: Done in 61ebe83. Extracted as a proper exported function `printEmpty(options, type, message)` in `src/lib/output.ts` rather than a local closure, so it could be reused. Now applied across inbox (3 sites), conversation/unread, conversation/helpers (`listConversationsWithUser`), workspace, users, and groups list — the latter four also had the same latent `--ndjson` stray-newline bug, now fixed.

src/commands/conversation/conversation.test.ts
scottlovegrove: Good catch — fixed in 61ebe83. The `--ndjson` empty-state tests in both `inbox.test.ts` and `conversation.test.ts` now assert `expect(logSpy).not.toHaveBeenCalled()`, so a regression that emits a blank line via `console.log('')` would be caught. Also added direct unit tests for `printEmpty` in `src/lib/output.test.ts` covering the same invariant.

src/lib/output.ts
```diff
@@ -227,6 +227,21 @@ export function printNdjson<T extends object>(items: T[], type?: EntityType, ful
     console.log(formatNdjson(items, type, full))
 }
 
+export function printEmpty(
+    options: { json?: boolean; ndjson?: boolean; full?: boolean },
+    type: EntityType,
+    message: string,
+): void {
```
scottlovegrove: We should use named params for this.

scottlovegrove: Done in e1df41d — `printEmpty` now takes a single named-params object `{ options, type, message }`. All eight call sites updated.

henningmu (commented):
✅

PR #206, opened 2026-05-08
feat(auth): support multiple authenticated Twist accounts

scottlovegrove:
## Summary

Adds multi-account support to twist-cli, mirroring [todoist-cli #300](https://github.com/Doist/todoist-cli/pull/300). The CLI can now hold credentials for multiple Twist accounts and target a specific one per command via `--user <id|email>`.

```bash
tw auth login                       # adds the account; first becomes default
tw account list                     # all stored accounts (with default marker)
tw account use <id|email>           # set the default account
tw --user <id|email> inbox          # one-off override for any command
tw auth logout --user <id|email>    # log out a specific account
```

Resolution order: `--user <ref>` > `account.defaultAccount` from config > the only stored account. Multiple accounts with no default makes commands error and ask for `--user` (or `tw account use`). `TWIST_API_TOKEN` still bypasses the resolver entirely.

## What's in here

- **Config schema v2** — `accounts: StoredAccount[]` + `account.defaultAccount`. `StoredAccount` carries `id`, `email`, `name`, `authMode`, `authScope`, optional `token` (plaintext fallback). `validateConfigForDoctor` updated to match.
- **Per-account keyring slots** — `accountForUser(id) → 'user-<id>'`. Legacy `'api-token'` slot is read once during migration then deleted.
- **One-time v1→v2 migration** in `src/lib/migrate-auth.ts`, run from `postinstall.ts`. Probes the legacy token via REST (no SDK import) to identify the user, moves the token to a per-account keyring slot, writes the v2 config, cleans up legacy state. Best-effort: any failure leaves v1 state untouched so the runtime fallback in `resolveActiveAccount` keeps serving the legacy token.
- **`--user` global flag** parsed in `src/lib/global-args.ts` (with `stripUserFlag` to keep commander happy since it has no global-option attachment) and validated in `src/index.ts` (rejects bare/empty/looks-like-subcommand values).
- **`tw account` command group** (`list`, `use`) under `src/commands/account/`. Plain output shows `Name <email> (id:N) (default)`; `--json` includes `name`, `isDefault`, `authMode`, `authScope`, `storage`.
- **`tw auth status`** now shows the default marker for the active account, lists other stored accounts, and points at `tw account use` / `--user`. Same JSON shape (`isDefault`, `storedAccounts[]`).
- **`tw auth login`** identifies the user behind the new token before persisting (so it lands in the right slot).
- **`tw auth token`** also probes the API to identify the user before saving.
- **`tw auth logout`** accepts `--user <ref>` to target a specific account.
- **OAuth login hang fix** (`cbc9678`) — `server.close()` waits for keep-alive sockets to drain, which kept the CLI hung for ~60s after a successful login. Force-terminate via `server.closeAllConnections()` and send `Connection: close` on responses.

## What's deliberately not ported

The todoist `auth_flags` / `--additional-scopes` system (app-management, backups) doesn't apply here — Twist's OAuth model has fixed `READ_ONLY_SCOPES` / `READ_WRITE_SCOPES` with no opt-in add-on scopes.

## Test plan

- [x] `npm run type-check`
- [x] `npx vitest run` (622 tests pass)
- [x] Manual: `tw auth login` for a second account — switch via `tw account use`, override per-command via `tw --user <ref> inbox`, log out a specific account via `tw auth logout --user <ref>`
- [x] Manual: confirm OAuth login no longer hangs after success page
- [ ] Manual: `tw doctor` against a fresh v2 config and against a v1 config requiring migration

### DIFF
--- a/src/commands/account/list.ts
+++ b/src/commands/account/list.ts
@@ -1,0 +1,46 @@
+import chalk from 'chalk'
+import { getDefaultAccountId } from '../../lib/accounts.js'
+import { listStoredAccounts } from '../../lib/auth.js'
+import { getConfig } from '../../lib/config.js'
+import { isAccessible } from '../../lib/global-args.js'
+import { formatJson } from '../../lib/output.js'
+
+export interface ListAccountsOptions {
+    json?: boolean
+}
+
+export async function listAccountsCommand(options: ListAccountsOptions): Promise<void> {
+    const accounts = await listStoredAccounts()
+    const config = await getConfig()
+    const defaultId = getDefaultAccountId(config)
+
+    if (options.json) {
+        console.log(
+            formatJson(
+                accounts.map((a) => ({
+                    id: a.id,
+                    email: a.email,
+                    name: a.name,
+                    isDefault: a.id === defaultId,
+                    authMode: a.authMode,
+                    authScope: a.authScope,
+                    storage: a.token ? 'config-file' : 'secure-store',
+                })),
+            ),
+        )
+        return
+    }
+
+    if (accounts.length === 0) {
+        console.log(chalk.dim('No stored Twist accounts. Run `tw auth login` to add one.'))
+        return
+    }
+
+    const accessible = isAccessible()
+    for (const a of accounts) {
+        const isDefault = a.id === defaultId
+        const marker = isDefault ? (accessible ? ' [default]' : chalk.green(' (default)')) : ''
+        const label = a.name ? `${a.name} <${a.email}>` : a.email
+        console.log(`${label} ${chalk.dim(`(id:${a.id})`)}${marker}`)
+    }
+}
--- a/src/commands/account/list.ts
+++ b/src/commands/account/list.ts
@@ -1,6 +1,5 @@
 import chalk from 'chalk'
-import { getDefaultAccountId } from '../../lib/accounts.js'
-import { listStoredAccounts } from '../../lib/auth.js'
+import { getDefaultAccountId, getStoredAccounts } from '../../lib/accounts.js'
 import { getConfig } from '../../lib/config.js'
 import { isAccessible } from '../../lib/global-args.js'
 import { formatJson } from '../../lib/output.js'
@@ -10,8 +9,9 @@ export interface ListAccountsOptions {
 }
 
 export async function listAccountsCommand(options: ListAccountsOptions): Promise<void> {
-    const accounts = await listStoredAccounts()
+    // One config read drives both the account list and the default lookup.
     const config = await getConfig()
+    const accounts = getStoredAccounts(config)
     const defaultId = getDefaultAccountId(config)
 
     if (options.json) {
--- a/src/commands/account/use.ts
+++ b/src/commands/account/use.ts
@@ -1,0 +1,20 @@
+import chalk from 'chalk'
+import { requireAccountByRef } from '../../lib/accounts.js'
+import { setDefaultAccountId } from '../../lib/auth.js'
+import { getConfig } from '../../lib/config.js'
+import { CliError } from '../../lib/errors.js'
+
+export async function useAccountCommand(ref: string | undefined): Promise<void> {
+    if (!ref) {
+        throw new CliError(
+            'MISSING_REF',
+            'Please provide an account id or email: `tw account use <id|email>`',
+        )
+    }
+
+    const config = await getConfig()
+    const { account } = requireAccountByRef(config, ref)
+    await setDefaultAccountId(account.id)
+
+    console.log(chalk.green('✓'), `Default account set to ${account.email} (id:${account.id})`)
+}
--- a/src/commands/account/use.ts
+++ b/src/commands/account/use.ts
@@ -1,7 +1,5 @@
 import chalk from 'chalk'
-import { requireAccountByRef } from '../../lib/accounts.js'
 import { setDefaultAccountId } from '../../lib/auth.js'
-import { getConfig } from '../../lib/config.js'
 import { CliError } from '../../lib/errors.js'
 
 export async function useAccountCommand(ref: string | undefined): Promise<void> {
@@ -12,9 +10,6 @@ export async function useAccountCommand(ref: string | undefined): Promise<void> 
         )
     }
 
-    const config = await getConfig()
-    const { account } = requireAccountByRef(config, ref)
-    await setDefaultAccountId(account.id)
-
+    const account = await setDefaultAccountId(ref)
     console.log(chalk.green('✓'), `Default account set to ${account.email} (id:${account.id})`)
 }
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -1,20 +1,31 @@
 import { Command } from 'commander'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-// Mock the auth module
+// Mock auth module
 vi.mock('../../lib/auth.js', async (importOriginal) => {
     const actual = await importOriginal<typeof import('../../lib/auth.js')>()
     return {
         ...actual,
-        saveApiToken: vi.fn(),
+        upsertAccount: vi.fn(),
         clearApiToken: vi.fn(),
         getAuthMetadata: vi.fn(),
+        listStoredAccounts: vi.fn(),
     }
 })
 
-// Mock the api module
+// Mock config (status reads it for default-account marker)
+vi.mock('../../lib/config.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('../../lib/config.js')>()
+    return {
+        ...actual,
+        getConfig: vi.fn(),
+    }
+})
+
+// Mock api module — getSessionUser used by status, createWrappedTwistClient used by login/token
 vi.mock('../../lib/api.js', () => ({
     getSessionUser: vi.fn(),
+    createWrappedTwistClient: vi.fn(),
 }))
 
 // Mock OAuth modules
@@ -38,12 +49,10 @@ vi.mock('../../lib/pkce.js', () => ({
     generateState: vi.fn(),
 }))
 
-// Mock open package
 vi.mock('open', () => ({
     default: vi.fn(),
 }))
 
-// Mock readline for interactive token input
 vi.mock('node:readline', () => ({
     createInterface: vi.fn(() => {
         const rl = {
@@ -54,14 +63,19 @@ vi.mock('node:readline', () => ({
     }),
 }))
 
-// Mock chalk to avoid colors in tests
 vi.mock('chalk')
 
 import { createInterface, type Interface } from 'node:readline'
 import { type User } from '@doist/twist-sdk'
 import open from 'open'
-import { getSessionUser } from '../../lib/api.js'
-import { clearApiToken, getAuthMetadata, saveApiToken } from '../../lib/auth.js'
+import { createWrappedTwistClient, getSessionUser } from '../../lib/api.js'
+import {
+    clearApiToken,
+    getAuthMetadata,
+    listStoredAccounts,
+    upsertAccount,
+} from '../../lib/auth.js'
+import { getConfig } from '../../lib/config.js'
 import { startCallbackServer } from '../../lib/oauth-server.js'
 import {
     buildAuthorizationUrl,
@@ -72,13 +86,14 @@ import { generateCodeChallenge, generateCodeVerifier, generateState } from '../.
 import { registerAuthCommand } from './index.js'
 
 const mockCreateInterface = vi.mocked(createInterface)
-
-const mockSaveApiToken = vi.mocked(saveApiToken)
+const mockUpsertAccount = vi.mocked(upsertAccount)
 const mockClearApiToken = vi.mocked(clearApiToken)
 const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
+const mockListStoredAccounts = vi.mocked(listStoredAccounts)
+const mockGetConfig = vi.mocked(getConfig)
 const mockGetSessionUser = vi.mocked(getSessionUser)
+const mockCreateWrappedTwistClient = vi.mocked(createWrappedTwistClient)
 
-// OAuth mocks
 const mockGenerateCodeVerifier = vi.mocked(generateCodeVerifier)
 const mockGenerateCodeChallenge = vi.mocked(generateCodeChallenge)
 const mockGenerateState = vi.mocked(generateState)
@@ -87,6 +102,24 @@ const mockStartCallbackServer = vi.mocked(startCallbackServer)
 const mockExchangeCodeForToken = vi.mocked(exchangeCodeForToken)
 const mockRegisterDynamicClient = vi.mocked(registerDynamicClient)
 const mockOpen = vi.mocked(open)
+
+const TEST_USER: User = {
+    id: 12345,
+    name: 'Scott',
+    shortName: 'scott',
+    bot: false,
+    timezone: 'UTC',
+    removed: false,
+    email: 'scott@example.com',
+    lang: 'en',
+}
+
+function stubProbeApiForUser(user: User = TEST_USER) {
+    mockCreateWrappedTwistClient.mockReturnValue({
+        users: { getSessionUser: vi.fn().mockResolvedValue(user) },
+        // biome-ignore lint/suspicious/noExplicitAny: minimal mock surface
+    } as any)
+}
 
 function createProgram() {
     const program = new Command()
@@ -101,10 +134,10 @@ describe('auth command', () => {
 
     beforeEach(() => {
         vi.clearAllMocks()
-
-        // Mock console.log to capture output
         consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
         errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+        mockListStoredAccounts.mockResolvedValue([])
+        mockGetConfig.mockResolvedValue({})
     })
 
     afterEach(() => {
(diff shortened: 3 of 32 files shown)


doistbot (commented):
This PR introduces solid multi-account support for the CLI, bringing welcome flexibility with the new `--user` targeting and a thoughtful v1-to-v2 configuration migration. The structural updates provide a great foundation for managing credentials, though there are opportunities to tighten architectural boundaries and optimize state lookups. A few refinements are needed around centralizing the `--user` flag parsing, consolidating redundant configuration reads, addressing edge cases in the legacy fallback and error handling paths, and ensuring proper test coverage for the new validation logic.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F206&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A99e682a0-4b00-11f1-873e-3f6e5291c545%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A99e682a0-4b00-11f1-873e-3f6e5291c545&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/token.ts
```diff
@@ -44,7 +45,22 @@ export async function loginWithToken(token?: string): Promise<void> {
             ])
         }
     }
-    const saveResult = await saveApiToken(token.trim(), { authMode: 'unknown' })
-    console.log(chalk.green('✓'), 'API token saved successfully!')
+    const trimmed = token.trim()
+
+    // Identify the Twist user behind the token so it lands in the right slot.
```
doistbot, line 51: [P2] This new probe path lets `users.getSessionUser()` errors escape as non-`CliError`s. For `tw auth token <bad-or-expired-token>` (or a transient network failure), the global handler will now print a raw stack/message instead of a structured auth error. Wrap the probe in a `CliError` (for example mapping 401s to an invalid-token/authentication message) before calling `upsertAccount()`.

src/index.ts
```diff
@@ -111,6 +119,38 @@ for (const [name, [description]] of Object.entries(commands)) {
...
+//
+// Commander has no global-option attachment, so leaving `--user` in argv
+// would make every subcommand error on it. We capture the value via
+// `parseGlobalArgs` first and remove it here. Before stripping we surface
+// usage errors that the parser can detect now but commander never will
+// because it never sees the flag — bare `--user`, `--user=` (empty), and
+// `--user <known-subcommand>` (almost always a forgotten value).
+{
+    const originalArgs = process.argv.slice(2)
+    const sawUserFlag = originalArgs.some((a) => a === '--user' || a.startsWith('--user='))
```
doistbot: [P2] This scan treats any `--user` token anywhere in `argv` as a global flag, even after a `--` terminator. That means commands that intentionally pass literal content like `--user=foo` after `--` can be misparsed as account selection and fail with the wrong account/error. Make this pre-parse validation honor `--` the same way `stripUserFlag()` does.

src/lib/auth.ts
```diff
@@ -19,282 +42,487 @@ export class NoTokenError extends CliError {
...
+    const config = await getConfig()
+    const accounts = getStoredAccounts(config)
+    const requestedRef = opts.ref ?? getRequestedUserRef()
+
+    // No accounts stored yet — fall through to legacy logout only on a
+    // v1-shaped config (no `accounts` key at all). Empty `accounts: []` is
+    // an already-clean v2 install; treat as a no-op rather than poking the
+    // legacy keyring.
+    if (accounts.length === 0) {
+        if (!Array.isArray(config.accounts)) {
```
doistbot: [P1] In the legacy-auth fallback, `clearApiToken()` ignores a provided `--user` ref and logs out the legacy credential unconditionally. On an unmigrated install, `tw auth logout --user wrong@example.com` will still remove the only saved token, which violates the new exact-match targeting contract. Reject `requestedRef` in this branch (or resolve the legacy token owner first) instead of calling `clearLegacyToken(config)` blindly.

src/lib/auth.ts
```diff
@@ -19,282 +42,487 @@ export class NoTokenError extends CliError {
...
-                'Token was migrated to secure storage,',
-            )
-            if (cleanupWarning) {
-                warn(cleanupWarning)
-            }
-        } catch (error) {
-            if (!(error instanceof SecureStoreUnavailableError)) {
-                throw error
+    let target: StoredAccount
+    if (requestedRef) {
```
doistbot, line 168: [P2] The logic for resolving the target account here duplicates the fallback logic in `clearApiToken` (lines 323-335). Additionally, both blocks manually reimplement the `requireAccountByRef` helper, and this block fails to reuse `getDefaultAccount(config)`. Consider extracting a shared `resolveTargetAccount(config, requestedRef)` helper to deduplicate this lookup and fallback logic.

src/index.ts
doistbot: [P2] This block re-implements the `--user` flag grammar that `parseGlobalArgs()` and `stripUserFlag()` already know about (`--user`, `--user=…`, missing value, subcommand-looking value). That leaves the parser, validator, and stripper split across multiple places. Please move the validation result into `global-args.ts` so all `--user` parsing rules have a single source of truth.

src/commands/account/use.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export async function useAccountCommand(ref: string | undefined): Promise<void> {
+    if (!ref) {
+        throw new CliError(
+            'MISSING_REF',
+            'Please provide an account id or email: `tw account use <id|email>`',
+        )
+    }
+
+    const config = await getConfig()
```
doistbot: [P2] This command resolves the ref here, and then `setDefaultAccountId()` immediately re-reads config and resolves it again. Since the first lookup is only needed for the success message, consider having `setDefaultAccountId()` accept the ref or return the resolved account so `tw account use` can reuse the same path instead of duplicating the lookup.

src/lib/auth.ts
```diff
@@ -19,282 +42,487 @@ export class NoTokenError extends CliError {
...
+    next = stripLegacyAuthFields(next)
+
+    try {
+        await setConfig(next)
+    } catch (error) {
+        // Config write is the source of truth — without it, later commands
+        // can't resolve the account even though the keyring holds the
+        // secret. Roll back the keyring so we don't leak a credential for
+        // a non-stored account, then surface the failure.
+        if (storedSecurely) {
```
doistbot: [P2] When a config write fails, this rollback deletes the token from the keyring. While this correctly prevents orphaning credentials for brand-new accounts, if the account already existed (`previouslyExisted === true`), this rollback destroys the user's previously working authentication. Since the unmodified config still points to `user-<id>`, leaving the new token in the keyring provides a much better degradation path where the CLI remains authenticated. Consider changing this to `if (storedSecurely && !previouslyExisted)`.

src/lib/accounts.ts
```diff
@@ -0,0 +1,117 @@
...
+
+export interface FindAccountResult {
+    account: StoredAccount
+    index: number
+}
+
+export class AccountNotFoundError extends CliError {
+    constructor(ref: AccountRef) {
+        super(
+            'ACCOUNT_NOT_FOUND',
```
doistbot, line 18: [P2] These new error codes are being introduced via the open-ended `ErrorCode` escape hatch instead of being registered in `src/lib/errors.ts`. That gives up the typed error-code registry/autocomplete the repo relies on. Please add `ACCOUNT_NOT_FOUND`/`NO_ACCOUNT_SELECTED` (and the other new codes from this PR) to the `ErrorCode` union.

src/lib/auth.ts
```diff
@@ -19,282 +42,487 @@ export class NoTokenError extends CliError {
...
+            authMode: 'unknown',
+            source: 'env',
+        }
     }
 
     const config = await getConfig()
-    const configToken = getConfigToken(config)
-    const secureStore = createSecureStore()
+    const accounts = getStoredAccounts(config)
+    const requestedRef = opts.ref ?? getRequestedUserRef()
```
doistbot: [P2] This makes `lib/auth` implicitly depend on CLI argv parsing, which breaks the existing layer boundary between command parsing and library logic. `resolveActiveAccount()` already accepts `{ ref?: string }`; prefer keeping `--user` resolution in `src/index.ts`/command handlers and passing the ref explicitly into the auth APIs instead of reading global process state here.

src/lib/auth.ts
```diff
@@ -19,282 +42,487 @@ export class NoTokenError extends CliError {
...
+export interface AuthMetadata {
+    authMode: AuthMode
     authScope?: string
+    source: 'env' | 'secure-store' | 'config-file'
+    accountId?: string
+    email?: string
 }
 
-export interface AuthMetadata {
+export interface ResolvedAccount {
```
doistbot, line 69: [P2] `ResolvedAccount` now has to represent both real stored accounts and synthetic env/legacy identities, but it does that with magic IDs (`'env'`, `'legacy'`) and empty-string emails. That makes the type misleading and forces downstream string checks like `resolved.id === 'env'`. Please model this as a discriminated union keyed by `source`, or make `id`/`email` optional for non-stored identities.

src/commands/account/list.ts
```diff
@@ -0,0 +1,46 @@
...
+import { getConfig } from '../../lib/config.js'
+import { isAccessible } from '../../lib/global-args.js'
+import { formatJson } from '../../lib/output.js'
+
+export interface ListAccountsOptions {
+    json?: boolean
+}
+
+export async function listAccountsCommand(options: ListAccountsOptions): Promise<void> {
+    const accounts = await listStoredAccounts()
```
doistbot: [P2] `listStoredAccounts()` already calls `getConfig()`, so the new `getConfig()` here rereads and reparses the same config file on every `tw account list`. Since both `accounts` and `defaultId` come from the same config snapshot, load config once and derive both values from it (`getStoredAccounts(config)` + `getDefaultAccountId(config)`) to avoid the duplicate FS read.

src/commands/auth/status.ts
```diff
@@ -19,21 +21,66 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
         throw error
     }
 
+    const metadata = await getAuthMetadata()
```
doistbot: [P2] This adds a full second auth/config resolution pass to `tw auth status`: `getSessionUser()` already goes through `getTwistClient()` -> `resolveActiveAccount()`, then `getAuthMetadata()` resolves the active account again, and `getConfig()` plus `listStoredAccounts()` each reread the config file (`listStoredAccounts()` calls `getConfig()` internally). On a cold invocation that means extra token-resolution work and two extra config reads after the API call. Consider resolving auth/config once here (or exposing a helper that returns metadata + accounts from one config snapshot) and batching any remaining independent reads.

src/commands/auth/auth.test.ts
```diff
@@ -270,295 +225,190 @@ describe('auth command', () => {
...
-            await expect(result).rejects.toHaveProperty('code', 'AUTH_FAILED')
-            await expect(result).rejects.toHaveProperty('hints')
+            expect(consoleSpy).toHaveBeenCalledWith(
+                '✓',
+                'Updated credentials for scott@example.com',
+            )
         })
 
-        it('handles token exchange errors', async () => {
+        it('cleanup runs on callback server error', async () => {
```
doistbot, line 316: [P2] This test no longer covers the cleanup guarantee from `loginWithOAuth`'s `finally` block. Because `startCallbackServer()` rejects here, no `cleanup` function is ever created, so the assertion only re-checks `AUTH_FAILED`. The more regression-prone path is: callback server starts successfully, then `exchangeCodeForToken()` / user probing / `upsertAccount()` fails and the returned `cleanup` must still run. Please restore one post-start failure case and assert that `cleanup()` is called.

src/index.ts
```diff
@@ -111,6 +119,38 @@ for (const [name, [description]] of Object.entries(commands)) {
     }
 }
 
+// Validate `--user` and strip it from argv before commander parses.
```
doistbot, line 136: [P2] The new `--user` pre-Commander validation/stripping block has no direct coverage. The helper tests in `global-args.test.ts` only exercise `parseGlobalArgs()`/`stripUserFlag()`, so regressions here (bare `--user`, `--user=`, subcommand-like values, JSON error formatting, or forgetting to rewrite `process.argv` before Commander parses) would still pass. Please add a thin index-level test around this block.

src/index.ts
```diff
@@ -111,6 +119,38 @@ for (const [name, [description]] of Object.entries(commands)) {
...
+// `--user <known-subcommand>` (almost always a forgotten value).
+{
+    const originalArgs = process.argv.slice(2)
+    const sawUserFlag = originalArgs.some((a) => a === '--user' || a.startsWith('--user='))
+    if (sawUserFlag) {
+        const ref = getRequestedUserRef()
+        const reportUserFlagError = (message: string, hints: string[]): never => {
+            const err = new CliError('USER_FLAG_INVALID', message, hints)
+            console.error(isJsonMode() ? formatErrorJson(err) : formatError(err))
+            process.exit(1)
```
doistbot: [P2] This reintroduces the `console.error(...) + process.exit(1)` pattern that the repo guidance explicitly asks us to avoid in favor of the shared `CliError` flow. Please throw the `CliError` here (or otherwise route it through the existing top-level `.catch`) so error formatting and cleanup stay centralized.

src/lib/auth.ts
scottlovegrove: Fixed in 772e62e (P1). clearApiToken now refuses `--user <ref>` on a v1-shaped install — throws AccountNotFoundError instead of blindly clearing the legacy credential.

src/index.ts
scottlovegrove: Fixed in 772e62e. parseGlobalArgs and the new validateUserFlag both honor `--` now (covered by a new 'ignores --user tokens after the -- terminator' test in global-args.test.ts).

src/lib/auth.ts
scottlovegrove: Done in 772e62e. Extracted `resolveTargetAccount(config, ref)` shared by resolveActiveAccount + clearApiToken; reuses getDefaultAccount and findAccountByRef.

src/commands/auth/status.ts
scottlovegrove: Done in 772e62e. tw auth status now does Promise.all([getAuthMetadata(), getConfig()]) and derives stored accounts + default from that single config snapshot. Removed the listStoredAccounts call entirely (was a third config read).

src/commands/auth/auth.test.ts
scottlovegrove: Done in 772e62e. Added a new test 'cleanup still runs when a post-callback step fails' that resolves startCallbackServer (returning the cleanup fn) and then makes upsertAccount throw — asserts cleanup() is still invoked via the finally block.

src/lib/accounts.ts
scottlovegrove: Done in 772e62e. Added ACCOUNT_NOT_FOUND, NO_ACCOUNT_SELECTED, INVALID_ACCOUNT, CONFIG_WRITE_FAILED, MISSING_REF, USER_FLAG_INVALID to the ErrorCode union.

src/lib/auth.ts
scottlovegrove: Done in 772e62e. Rollback is now `if (storedSecurely && !previouslyExisted)` — re-auth of an existing account leaves the working secret intact when the config write fails.

src/commands/auth/token.ts
scottlovegrove: Wrapped in 772e62e: 401 → INVALID_TOKEN, network/other → AUTH_FAILED. No more raw stack.

src/commands/account/list.ts
scottlovegrove: Done in 772e62e. tw account list reads getConfig once and derives both via getStoredAccounts(config) + getDefaultAccountId(config) — one FS read.

src/index.ts
scottlovegrove: Done in 772e62e. Added a validateUserFlag describe block in global-args.test.ts covering bare --user, --user= (empty), --user --json, --user <known-subcommand>, and post-`--` tokens.

src/index.ts
scottlovegrove: Done in 772e62e. Extracted reportError() that the bottom .catch and the pre-parse path both call, so format/exit logic is centralized. The pre-parse branch still needs an explicit process.exit because it runs synchronously before commander registers anything (the bottom .catch can't see it); commented inline.

src/index.ts
scottlovegrove: Done in 772e62e. `validateUserFlag(argv, knownCommands)` lives in global-args.ts; index.ts is now a thin caller. All --user grammar (parse, validate, strip) is in one module.

src/commands/account/use.ts
scottlovegrove: Done in 772e62e. setDefaultAccountId now takes a ref and returns the resolved StoredAccount; `tw account use` reads it and skips the redundant lookup.

src/lib/auth.ts
scottlovegrove: Done in 772e62e. Dropped the magic 'env'/'legacy' ids; ResolvedAccount.id and .email are now optional, and legacy fallback carries an explicit `legacy: true` flag. Callers branch on `source` / `legacy` instead of sentinel strings.

src/lib/auth.ts
scottlovegrove: Done in 772e62e. Removed the getRequestedUserRef import from lib/auth.ts. Callers (api.ts getTwistClient/getCurrentWorkspaceId/setActiveAccountWorkspace, commands/auth/logout.ts) now read it themselves and pass via opts.ref.

pedroalves0:
_re-rolling the review assignment as I'll be OOO tomorrow._

scottlovegrove:
Closing in favour of #216 — re-implementation of this branch on top of `main` after #215 landed (cli-core auth). The original branch can't be mechanically rebased (`oauth-server.ts` was deleted, `auth.ts` was rewritten), so #216 starts fresh while preserving the same feature surface. The `~60s hang fix` from `cbc9678` is already in cli-core (`closeAllConnections` in `flow.js`).

PR #208, opened 2026-05-09
refactor: Start integrating `@doist/cli-core`

scottlovegrove:
## Summary

First consumer-side integration of [@doist/cli-core](https://www.npmjs.com/package/@doist/cli-core) 0.9.0 into twist-cli. Mirrors todoist-cli [#313](https://github.com/Doist/todoist-cli/pull/313) (errors / config / spinner helpers / global-args / output) and [#314](https://github.com/Doist/todoist-cli/pull/314) (full spinner delegation via `createSpinner`) as a single combined PR.

The local config API stays at `getConfig` / `setConfig` / `updateConfig` — they're now thin wrappers around cli-core's `readConfig` / `writeConfig` / `updateConfig`. Zero call-site churn outside `src/lib/config.ts` itself.

## What's wired through cli-core now

| Module | Was | Now |
| --- | --- | --- |
| `src/lib/errors.ts` | full local `CliError` class | local subclass over cli-core's `CliError<ErrorCode>` — keeps the historical positional `(code, message, hints?, type?)` signature so the ~hundreds of existing call sites stay untouched. `code` accepts `ErrorCode \| CliErrorCode` so cli-core's canonical codes (`CONFIG_*`, etc.) type-check at call sites without redeclaring them locally. |
| `src/lib/config.ts` | local path / read / strict-read / write impls using `node:fs/promises`, `node:os`, `node:path` | thin wrappers: `getConfigPath` (lazy, takes `'twist-cli'`), `getConfig` → `readConfig`, `setConfig` → `writeConfig`, `updateConfig` → `updateConfig`. Schema (`Config`, `AuthMode`, `UpdateChannel`, `UserSettings`, the `KNOWN_*_KEYS` sets, `validateConfigForDoctor`) stays. `readConfigStrict` translates cli-core's discriminated `ReadConfigStrictResult` into the same three twist `CliError`s with identical hint copy. |
| `src/lib/spinner.ts` | local 130-line impl with yocto-spinner, early-spinner singleton, stdout interception, adoption/release semantics, `LoadingSpinner` class, colour palette | `const spinner = createSpinner({ isDisabled: shouldDisableSpinner })` + destructured exports. The CLI-specific bit (`TW_SPINNER` opt-out + `--json` / `--ndjson` / `--no-spinner` fan-out) is wired in as `isDisabled`. `yocto-spinner` drops from direct deps (still resolved transitively via cli-core). |
| `src/lib/global-args.ts` | `process.env.CI` direct read | `isCI()` (also opt-out aware: `CI='false'` is treated as not-CI) |
| `src/lib/output.ts` | inline `JSON.stringify(_, null, 2)` / `items.map(JSON.stringify).join('\n')` | delegates serialization to `formatJson` / `formatNdjson`; entity-aware processing via `filterEntityFields` and the `_meta` cursor sentinel for paginated NDJSON stays. Paginated formatters route through the local helpers (single serialization path). |

`formatErrorJson` deliberately keeps its inline compact `JSON.stringify` — it emits single-line JSON, not pretty-print, so it doesn't fit cli-core's `formatJson` primitive. `shouldDisableSpinner` likewise stays local since it bundles the `TW_SPINNER` env var with the global-args flag fan-out.

## Behaviour change worth flagging

`yocto-spinner.success()` and `.error()` already prepend their own `✔` / `✖` glyphs. The old code added a manual `✓ ` / `✗ ` on top, which produced doubled symbols in real terminals (e.g. `✔ ✓ Done`). cli-core 0.3+ drops the manual prefixes — so the success / fail line now renders as a single `✔ Done` instead of `✔ ✓ Done`. Surfaced upstream during the cli-core review and confirmed in cli-core's own spinner tests.

`updateConfig` now uses cli-core's strict-read semantics: a broken config file throws instead of being silently overwritten (the previous lenient `getConfig` + `setConfig` composition could clobber data the user might have recovered by hand-fixing the file).

## vitest config

Adds `server.deps.inline: ['@doist/cli-core']` to `vitest.config.ts`. Without it, vitest treats the real-installed cli-core as external and `vi.mock('@doist/cli-core', …)` / `vi.doMock('node:fs/promises', …)` don't reach its compiled imports, breaking the `auth` / `config` suites.

## Tests

`src/lib/spinner.test.ts` is trimmed to a single smoke test that mocks `@doist/cli-core#createSpinner` and asserts the kit is built with `{ isDisabled: shouldDisableSpinner }`. The previous 333 lines of behavioural coverage are now in cli-core's own spinner suite; the wiring assertion is the one thing that has to live here.

`src/lib/config.test.ts` gains two new blocks:
- `readConfigStrict wrapper` — covers all five state→`CliError` translations (missing, present, read-failed, invalid-json, invalid-shape).
- `thin config wrappers` — proves `getConfigPath`, `getConfig`, `setConfig`, `updateConfig` each forward to cli-core with the resolved `'twist-cli'` app-name path. A wrong app name would silently redirect every config read/write — these tests are the tripwire.

`src/lib/global-args.test.ts` adds a regression locking in cli-core's `isCI()` opt-out: `CI='false'` no longer disables the spinner.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check` — 0 warnings, 0 errors
- [x] `npm test` — **579 / 579** pass
- [x] `npm run build`
- [x] `npm ls yocto-spinner` confirms cli-core still resolves it transitively
- [ ] Manually exercise `tw auth status`, `tw config view`, `tw doctor` against a real config to confirm parity at the wrapper boundary
- [ ] Run a long-running command (`tw inbox`) in a TTY to confirm the spinner still appears, the early spinner is adopted, and **no doubled `✔ ✓` glyphs** on success
- [ ] `CI=1 tw inbox`, `tw inbox --json`, `TW_SPINNER=false tw inbox` to confirm spinner suppression via `isCI()` and `shouldDisableSpinner`


src/lib/config.ts
```diff
@@ -1,9 +1,22 @@
-import { chmod, mkdir, readFile, writeFile } from 'node:fs/promises'
-import { homedir } from 'node:os'
-import { dirname, join } from 'node:path'
+import {
+    getConfigPath as getConfigPathCore,
+    readConfig as readConfigCore,
+    readConfigStrict as readConfigStrictCore,
+    writeConfig as writeConfigCore,
+} from '@doist/cli-core'
```
doistbot, line 7: [P2] The PR description mentions that the local `updateConfig` is now a thin wrapper around `cli-core`'s `updateConfig`. However, `updateConfig` isn't imported here, and the local implementation was left unchanged, missing the delegation to `cli-core`.

src/lib/output.ts
```diff
@@ -164,7 +159,7 @@ export function formatPaginatedJson<T extends object>(
     full = false,
 ): string {
     const results = type ? filterEntityFields(data.results, type, full) : data.results
-    return JSON.stringify({ results, nextCursor: data.nextCursor }, null, 2)
+    return formatJsonCore({ results, nextCursor: data.nextCursor })
```
doistbot: [P2] `formatPaginatedJson` (and the NDJSON variant just below) now call the cli-core serializers directly instead of reusing the local `formatJson`/`formatNdjson` wrappers introduced above. That leaves a second serialization path to keep in sync. After computing `results`, route through the local helpers so paginated output inherits any future formatting changes automatically.

src/lib/output.ts
```diff
@@ -173,11 +168,8 @@ export function formatPaginatedNdjson<T extends object>(
     full = false,
 ): string {
     const results = type ? filterEntityFields(data.results, type, full) : data.results
-    const lines = results.map((item) => JSON.stringify(item))
-    if (data.nextCursor) {
-        lines.push(JSON.stringify({ _meta: true, nextCursor: data.nextCursor }))
-    }
-    return lines.join('\n')
+    if (!data.nextCursor) return formatNdjsonCore(results)
+    return formatNdjsonCore([...results, { _meta: true, nextCursor: data.nextCursor }])
```
doistbot: [P2] When `nextCursor` is present, this spreads `results` into a new array before serializing, which adds an extra O(n) copy and extra transient memory for every paginated NDJSON response. The previous implementation appended the `_meta` line without duplicating the whole page. Please keep the trailer append behavior (or teach `formatNdjsonCore` about a trailer item) so large list outputs don’t pay for a full extra array allocation.

src/lib/spinner.test.ts
```diff
@@ -1,343 +0,0 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
```
doistbot, line 1: [P2] Removing this file drops the only runtime coverage that `spinner.ts` actually passes Twist's disable policy into the cli-core spinner. `global-args.test.ts` only checks `shouldDisableSpinner()` in isolation, so a broken `createSpinner({ isDisabled })` handoff or bad export destructuring would now pass CI. Please keep a small smoke test here by mocking `@doist/cli-core#createSpinner` and asserting the exported spinner helpers are created with `shouldDisableSpinner`.

src/lib/global-args.ts
```diff
@@ -110,7 +112,7 @@ export function includePrivateChannels(): boolean {
 
 export function shouldDisableSpinner(): boolean {
     if (process.env.TW_SPINNER === 'false') return true
-    if (process.env.CI) return true
+    if (isCI()) return true
```
doistbot, line 115: [P2] This switch to `isCI()` is meant to change the `CI='false'` case, but the updated tests still only cover `CI='true'`. Please add a regression test that `process.env.CI = 'false'` leaves `shouldDisableSpinner()` false, otherwise the new consumer-side behavior isn't verified.

src/lib/config.test.ts
```diff
@@ -40,3 +53,61 @@ describe('validateConfigForDoctor', () => {
         )
     })
 })
+
+describe('readConfigStrict wrapper', () => {
```
doistbot, line 77: [P2] This new suite exercises only `readConfigStrict`. The refactor also introduced `getConfig`, `setConfig`, and `getConfigPath` as cli-core passthroughs, and none of the tests assert that they use the resolved `twist-cli` path. A wrong app name or path-forwarding bug would redirect all config I/O without failing tests. Please add a small wrapper test around those thin read/write helpers.

doistbot (commented):
This PR cleanly integrates `@doist/cli-core` 0.9.0 into `twist-cli`, adopting the central utilities for errors, configuration, output formatting, and spinners while keeping existing call sites intact. Centralizing these foundations significantly reduces local boilerplate and aligns the project more closely with our standard tooling ecosystem. A few minor adjustments are needed to fully wire up the `updateConfig` delegate, streamline paginated JSON/NDJSON formatting to avoid unnecessary memory allocations, and ensure test coverage is preserved for the spinner, CI environment flags, and configuration path resolution.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F208&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae49b3ab0-4bdd-11f1-9230-225b72b559f3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae49b3ab0-4bdd-11f1-9230-225b72b559f3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/config.ts
scottlovegrove: Addressed in dfc5c5e — `updateConfig` now imports `updateConfig as updateConfigCore` from `@doist/cli-core` and the body is `await updateConfigCore<Config>(getConfigPath(), updates)`. Using cli-core's strict-read semantics is a small behavioural improvement: a broken config file now throws instead of being silently overwritten.

src/lib/output.ts
scottlovegrove: Addressed in dfc5c5e — both paginated formatters now route through the local `formatJson` / `formatNdjson` wrappers (which in turn delegate to cli-core), so there's a single serialization path.

src/lib/output.ts
scottlovegrove: Addressed in dfc5c5e — body is now `formatNdjson(data.results, type, full)` (single pass, no copy) and the cursor trailer is appended via string concat: `\`${body}\n${trailer}\``. Empty-results-with-cursor returns just the trailer. No more `[...results, …]` allocation.

src/lib/spinner.test.ts
scottlovegrove: Addressed in dfc5c5e — re-added `src/lib/spinner.test.ts` with a single smoke test that mocks `@doist/cli-core#createSpinner` and asserts it's called with `{ isDisabled: shouldDisableSpinner }`. A broken handoff or destructure would now fail this test.

src/lib/global-args.ts
scottlovegrove: Addressed in dfc5c5e — added "treats CI='false' as opt-out (does not disable the spinner)" in `global-args.test.ts`'s `shouldDisableSpinner` block. Locks in the cli-core `isCI()` opt-out behaviour that the swap unlocked.

src/lib/config.test.ts
scottlovegrove: Addressed in dfc5c5e — added a `thin config wrappers` block in `config.test.ts` that mocks all four cli-core functions and asserts `getConfigPath`, `getConfig`, `setConfig`, and `updateConfig` each forward to cli-core with the resolved `'twist-cli'` app-name path. A wrong app name now fails CI.

PR #210, opened 2026-05-09
refactor: delegate markdown + changelog to @doist/cli-core

scottlovegrove:
## Summary

Ports [todoist-cli#319](https://github.com/Doist/todoist-cli/pull/319) to twist-cli — delegates markdown rendering and the `changelog` command to `@doist/cli-core@0.9.0` (already pinned), and widens the top-level error handler so cli-core-thrown errors route through twist's formatter.

- **Markdown rendering** — `src/lib/markdown.ts` now delegates `preloadMarkdown` / `renderMarkdown` to `@doist/cli-core/markdown`, keeping the local `twist-mention://` → `@mention` preprocessor on top. Drops the local `marked-terminal-renderer` ambient `.d.ts` shim now that cli-core owns the typing.
- **Bootstrap preload** — `src/index.ts` kicks off `preloadMarkdown()` in parallel with the dynamic command-module load when the output mode renders markdown (skipped for `--json`/`--ndjson`/`--raw` and non-rendering commands like `changelog`/`update`/`completion`/`doctor`/`auth`/`config`/`skill`). The import cost overlaps with module load — zero added latency.
- **Changelog command** — `src/commands/changelog.ts` is now a thin wrapper over `registerChangelogCommand` from `@doist/cli-core/commands`, passing twist-specific options (`headingLevel: 'flexible'`, `continuationIndent: true`, `filterEmptyVersions: true`) so rendered output is unchanged. Local behavior tests deleted — cli-core covers them.
- **Top-level error handler** — `src/index.ts` now matches `instanceof BaseCliError` (cli-core class) so errors thrown from cli-core commands route through the existing `formatError` path. Local `CliError` extends base, so subclass throws still match. `formatError` / `formatErrorJson` parameter types widened to `BaseCliError<string>`.

Net diff: −301 / +114 lines.

## Test plan

- [x] `npm run lint:check` (oxlint + oxfmt)
- [x] `npm run type-check` (`tsc --noEmit`)
- [x] `npm test` — 583 tests pass
- [x] `npm run build`
- [x] Smoke `tw changelog -n 1` — output unchanged (heading green-bold, dimmed bullets, footer link with current version)
- [x] Smoke `tw changelog -n abc` — cli-core throws `CliError('INVALID_TYPE')`; twist's widened handler renders `Error: INVALID_TYPE\nCount must be a positive integer`, exit 1
- [ ] Smoke `tw thread view <id>` — markdown body renders with ANSI styling (not exercised in CI)
- [ ] Smoke `tw msg view <id> --json` — markdown rendering skipped, output unchanged


src/lib/markdown.ts
```diff
@@ -1,23 +1,21 @@
...
+export async function preloadMarkdown(): Promise<void> {
+    if (!preloadPromise) preloadPromise = corePreloadMarkdown()
+    return preloadPromise
+}
+
 export async function renderMarkdown(content: string): Promise<string> {
-    const processed = preprocessMentions(content)
-    const rendered = await getMarkedInstance().parse(processed)
-    return rendered.trimEnd()
+    if (!preloadPromise) return content
```
doistbot: [P2] Silent failure when not preloaded. Returning raw content tightly couples this utility function to the `needsMarkdown` heuristic in `src/index.ts`. If a command renders markdown but isn't included in the preload path (e.g. testing a sub-command in isolation, or a newly added command), it will silently fail to render rather than falling back to a lazy import. Ensure the function contract is always upheld by making it lazily initialize:
```typescript
    await preloadMarkdown()
    return coreRenderMarkdown(preprocessMentions(content))
```

src/index.ts
```diff
@@ -163,9 +164,28 @@ if (process.argv[2] === 'completion-server') {
...
+            'changelog',
+            'update',
+            'completion',
+            'doctor',
+            'auth',
+            'config',
+            'skill',
+        ])
+        const argv = process.argv.slice(2)
+        const wantsRaw = argv.includes('--raw') || argv.includes('--ndjson')
```
doistbot: [P2] Reinvents parsing for the `--ndjson` global flag via a brittle string search. Since the CLI already parses this robustly into the global args store, consider exporting an `isNdjsonMode()` helper from `src/lib/global-args.ts` (matching `isJsonMode()`) to reuse the cached parsed state.

src/lib/global-args.ts
```diff
@@ -1,102 +1,119 @@
...
+ */
+export function parseGlobalArgs(argv?: string[]): TwGlobalArgs {
+    const args = argv ?? process.argv
+    const base = parseCoreGlobalArgs(args)
+    const local = parseTwLocalFlags(args)
 
-function getGlobalArgs(): GlobalArgs {
-    if (!cached) {
-        cached = parseGlobalArgs()
+    return {
```
doistbot, line 112: [P2] Spreading `parseCoreGlobalArgs()` straight through here exposes cli-core's full global option shape in Twist's public `GlobalArgs` surface (the new tests already show `quiet`/`verbose`), but `src/index.ts` does not register those options with Commander. That creates a type/API leak where helpers believe the CLI supports flags the binary will actually reject. Please either add the matching top-level options or narrow the exported shape back down to Twist-supported flags before returning it.

src/lib/markdown.test.ts
```diff
@@ -1,16 +1,17 @@
-import { describe, expect, it } from 'vitest'
-import { renderMarkdown } from './markdown.js'
+import { beforeAll, describe, expect, it } from 'vitest'
+import { preloadMarkdown, renderMarkdown } from './markdown.js'
 
 describe('markdown', () => {
-    it('renders markdown via marked-terminal-renderer', async () => {
+    beforeAll(async () => {
```
doistbot: [P2] The optimization to skip markdown rendering when `preloadMarkdown()` hasn't been called (e.g., in `--json` mode) is currently untested. Because `preloadMarkdown()` is called in a top-level `beforeAll`, the `preloadPromise` singleton is permanently set for the entire file, masking the bypass behavior. Consider adding a test for the raw bypass before this block, and nesting the preloaded tests inside their own `describe` block.

src/commands/changelog.test.ts
```diff
@@ -2,168 +2,62 @@ import { Command } from 'commander'
...
     })
 
     afterEach(() => {
         vi.restoreAllMocks()
-        process.exitCode = undefined
+        mockReadFile.mockReset()
     })
 
-    it('shows last 5 versions by default', async () => {
+    it('passes the twist CHANGELOG.md path through to cli-core', async () => {
```
doistbot, line 54: [P2] These wrapper tests no longer cover the twist-owned behavior that matters most after the cli-core delegation: the specific options passed through (`headingLevel`, `continuationIndent`, `filterEmptyVersions`, repo/version). With the current fixture, regressions in those flags would still pass because there are no wrapped bullets or empty versions to expose them. Add a focused assertion on the options object (or fixtures that depend on those flags) so the wrapper contract is locked down.

src/index.ts
```diff
@@ -163,9 +164,28 @@ if (process.argv[2] === 'completion-server') {
...
+        const needsMarkdown = !noMarkdownCommands.has(commandName) && !isJsonMode() && !wantsRaw
+
         startEarlySpinner()
         try {
+            const markdownReady = needsMarkdown ? preloadMarkdown() : undefined
```
doistbot: [P2] If `preloadMarkdown()` rejects before `loader()` completes, it will cause an `unhandledRejection` that immediately crashes the process (leaving the early spinner stuck, violating the error handling guidelines). Consider using `Promise.all` to safely await both in parallel without unhandled rejections: `const [register] = await Promise.all([loader(), needsMarkdown ? preloadMarkdown() : undefined])`.

doistbot (commented):
This is a great refactor that successfully streamlines the CLI by delegating markdown rendering and the changelog command to `@doist/cli-core`. Relying on shared core utilities will significantly reduce maintenance overhead and improve long-term consistency. A few edge cases require attention, including safely handling parallel markdown preloading to prevent unhandled rejections, preventing unsupported core flags from leaking into the global arguments shape, improving the robustness of lazy markdown initialization and NDJSON parsing, and refining tests to properly cover Twist-specific changelog options and preload bypass behavior.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F210&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A13952280-4be9-11f1-9b3a-01698657e718%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A13952280-4be9-11f1-9b3a-01698657e718&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/markdown.ts
scottlovegrove: Fixed in 407352e — `renderMarkdown` now `await`s `preloadMarkdown()` internally, so it always upholds its contract regardless of whether the bootstrap path preloaded. The bootstrap preload in `src/index.ts` is now a pure perf optimization (parallelize the dynamic import with the command-module load); correctness no longer depends on the `needsMarkdown` heuristic.

src/index.ts
scottlovegrove: Fixed in 407352e — added `isNdjsonMode()` to `src/lib/global-args.ts` (matching `isJsonMode()`) and replaced the argv string-search in `src/index.ts` with it. The cached parsed state is the single source of truth.

src/lib/global-args.ts
scottlovegrove: Fixed in 407352e — narrowed the exported `TwGlobalArgs` to drop `quiet` / `verbose` (since twist does not register `--quiet` / `--verbose` with Commander). The internal store keeps the full cli-core shape via a private `FullArgs` type so `createGlobalArgsStore` / `createAccessibleGate` / `createSpinnerGate` still satisfy `T extends GlobalArgs`. Updated the default-shape fixture in `global-args.test.ts` to match.

src/lib/markdown.test.ts
scottlovegrove: Resolved in 407352e — once `renderMarkdown` self-inits, there is no bypass behavior left to test. Dropped the `beforeAll(preloadMarkdown)` since lazy init now handles it.

src/commands/changelog.test.ts
scottlovegrove: Fixed in 407352e — extended the wrapper-test fixture with both `# 1.x` and `## 1.x` headings, a deps-only release, and a wrapped continuation line. Added three focused assertions that lock down each twist-specific option (`headingLevel: flexible`, `filterEmptyVersions: true`, `continuationIndent: true`); a regression in any one would now fail the suite.

src/index.ts
scottlovegrove: Fixed in 407352e — switched to `const [register] = await Promise.all([loader(), needsMarkdown ? preloadMarkdown() : undefined])`. A preload rejection now co-rejects with the loader instead of leaking as an `unhandledRejection`, and the existing `try/catch` in the bootstrap routes the error through `stopEarlySpinner` + `formatError` as expected.

craigcarlyle:
✅

PR #211, opened 2026-05-10
refactor(config): migrate updateChannel → update_channel (read seam)

scottlovegrove:
## Summary

Prerequisite for a follow-up PR that delegates `tw update` to `@doist/cli-core/commands` (mirror of [todoist-cli#321](https://github.com/Doist/todoist-cli/pull/321)). cli-core's `registerUpdateCommand` reads/writes the config key `update_channel` (snake_case, hardcoded). Twist's schema is camelCase (`updateChannel`). This PR aligns the one field cli-core requires, with transparent migration of existing user configs.

- Rename `Config.updateChannel` → `Config.update_channel`. The only snake_case key in an otherwise camelCase schema; the inconsistency is intentionally scoped to this one field — alternative is a translation layer at every callsite.
- Add `migrateLegacyKeys` helper that renames `updateChannel` → `update_channel` at the read seam (`getConfig`, `readConfigStrict`). In-memory `Config` is always canonical; on-disk file gets rewritten on the next config update.
- `updateConfig` is now a read-merge-write wrapper rather than delegating to cli-core's atomic variant, so legacy keys are dropped from disk on the next mutation rather than lingering across updates.
- `validateConfigForDoctor` accepts both keys: `update_channel` validated as before; `updateChannel` emits a soft warning that it will be auto-migrated. Both stay in `KNOWN_CONFIG_KEYS` so doctor doesn't flag a legacy file as having "unrecognized" keys.

## Behaviour

- **User with new config (`update_channel`):** no change.
- **User with legacy config (`updateChannel`):** every read transparently renames in memory. First config write (`tw update switch`, `tw auth login`, etc.) persists the canonical key and drops the legacy one. No user action required.
- **\`tw doctor\` against legacy config:** emits a soft warning `"updateChannel is a legacy key — will be migrated to update_channel automatically on next config write"`. Still PASSes; doesn't error.

## Test plan

- [x] \`npm run type-check\` clean
- [x] \`npm run lint:check\` clean
- [x] \`npm test\` — 593 tests pass (was 586; +7 migration cases)
- [x] \`npm run build\`
- [x] Smoke: \`tw update --channel\` against default config reports \`stable\`
- [ ] Manual: write legacy \`{ "updateChannel": "pre-release" }\` to \`~/.config/twist-cli/config.json\`, run \`tw update --channel\` → reports \`pre-release\`. Run \`tw update switch --stable\` → on-disk file now has \`update_channel: 'stable'\`, no \`updateChannel\` key.
- [ ] Manual: \`tw doctor\` against the legacy config shows the soft warning but doesn't error.

### DIFF
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -130,7 +130,7 @@ describe('doctor command', () => {
                 update_channel: 'pre-release',
             }),
         )
-        mockGetConfig.mockResolvedValue({ update_channel: 'pre-release' })
+        mockGetConfig.mockResolvedValue({ updateChannel: 'pre-release' })
         mockProbeApiToken.mockResolvedValue({
             token: 'plaintext-token',
             metadata: { authMode: 'read-write', source: 'config-file' },
@@ -195,8 +195,59 @@ describe('doctor command', () => {
         expect(process.exitCode).toBeUndefined()
     })
 
+    it('reads legacy updateChannel from on-disk config and reports the channel', async () => {
+        // Disk still has the legacy camelCase key; read-seam translates it,
+        // so doctor reports the configured channel exactly as it would for
+        // a canonical `update_channel` file. No "unrecognized key" warning.
+        mockReadFile.mockResolvedValue(
+            JSON.stringify({
+                token: 'plaintext-token',
+                updateChannel: 'pre-release',
+            }),
+        )
+        mockGetConfig.mockResolvedValue({ updateChannel: 'pre-release' })
+        mockProbeApiToken.mockResolvedValue({
+            token: 'plaintext-token',
+            metadata: { authMode: 'read-write', source: 'config-file' },
+        })
+        mockFetch('1.0.0')
+
+        const program = createProgram()
+        await program.parseAsync(['node', 'tw', 'doctor'])
+
+        expect(consoleSpy).toHaveBeenCalledWith(
+            expect.stringContaining('PASS CLI is up to date on pre-release (v1.0.0)'),
+        )
+        const configWarning = consoleSpy.mock.calls.find(
+            (call: unknown[]) =>
+                typeof call[0] === 'string' &&
+                (call[0] as string).includes('WARN Config file is readable but'),
+        )?.[0]
+        // updateChannel is a known legacy key — must not show as unrecognized.
+        expect(configWarning ?? '').not.toContain('unrecognized key "updateChannel"')
+    })
+
+    it('flags invalid legacy updateChannel value in the validator', async () => {
+        mockReadFile.mockResolvedValue(
+            JSON.stringify({
+                updateChannel: 'beta',
+            }),
+        )
+        mockFetch('1.0.0')
+
+        const program = createProgram()
+        await program.parseAsync(['node', 'tw', 'doctor'])
+
+        const configWarning = consoleSpy.mock.calls.find(
+            (call: unknown[]) =>
+                typeof call[0] === 'string' &&
+                (call[0] as string).includes('WARN Config file is readable but'),
+        )?.[0]
+        expect(configWarning).toContain('updateChannel must be one of: stable, pre-release')
+    })
+
     it('normalizes invalid update channel values to stable', async () => {
-        mockGetConfig.mockResolvedValue({ update_channel: 'beta' as never })
+        mockGetConfig.mockResolvedValue({ updateChannel: 'beta' as never })
         mockFetch('1.0.0')
 
         const program = createProgram()
--- a/src/commands/doctor.test.ts
+++ b/src/commands/doctor.test.ts
@@ -127,10 +127,10 @@ describe('doctor command', () => {
         mockReadFile.mockResolvedValue(
             JSON.stringify({
                 token: 'plaintext-token',
-                updateChannel: 'pre-release',
+                update_channel: 'pre-release',
             }),
         )
-        mockGetConfig.mockResolvedValue({ updateChannel: 'pre-release' })
+        mockGetConfig.mockResolvedValue({ update_channel: 'pre-release' })
         mockProbeApiToken.mockResolvedValue({
             token: 'plaintext-token',
             metadata: { authMode: 'read-write', source: 'config-file' },
@@ -161,7 +161,7 @@ describe('doctor command', () => {
                 pendingSecureStoreClear: 'yes',
                 currentWorkspace: 'abc',
                 authMode: 'admin',
-                updateChannel: 'beta',
+                update_channel: 'beta',
                 extraSetting: true,
             }),
         )
@@ -184,7 +184,7 @@ describe('doctor command', () => {
         expect(configWarning).toContain('pendingSecureStoreClear must be a boolean')
         expect(configWarning).toContain('currentWorkspace must be a positive integer')
         expect(configWarning).toContain('authMode must be one of: read-only, read-write, unknown')
-        expect(configWarning).toContain('updateChannel must be one of: stable, pre-release')
+        expect(configWarning).toContain('update_channel must be one of: stable, pre-release')
         expect(consoleSpy).toHaveBeenCalledWith(
             expect.stringContaining('PASS Authenticated as person@example.com via secure-store'),
         )
@@ -196,7 +196,7 @@ describe('doctor command', () => {
     })
 
     it('normalizes invalid update channel values to stable', async () => {
-        mockGetConfig.mockResolvedValue({ updateChannel: 'beta' as never })
+        mockGetConfig.mockResolvedValue({ update_channel: 'beta' as never })
         mockFetch('1.0.0')
 
         const program = createProgram()
--- a/src/lib/config.ts
+++ b/src/lib/config.ts
@@ -2,6 +2,7 @@ import {
     getConfigPath as getConfigPathCore,
     readConfig as readConfigCore,
     readConfigStrict as readConfigStrictCore,
+    updateConfig as updateConfigCore,
     writeConfig as writeConfigCore,
 } from '@doist/cli-core'
 import { CliError } from './errors.js'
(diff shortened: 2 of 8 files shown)


src/lib/config.ts
```diff
@@ -50,18 +53,41 @@ export interface Config {
...
 
+/**
+ * Rename twist's legacy `updateChannel` field to the canonical
+ * `update_channel` (the shape cli-core's update command expects). Runs on
+ * every read so the in-memory `Config` is always canonical, even when the
+ * on-disk file is still in the old shape — the file is rewritten with the
+ * canonical key on the next `setConfig` / `updateConfig` call.
+ */
+function migrateLegacyKeys(raw: Record<string, unknown>): Record<string, unknown> {
+    if (!('updateChannel' in raw)) return raw
```
doistbot: [P2] If the config file is manually edited to contain `null` or a primitive value like a literal string (which are valid JSON), `raw` will not be an object and the `in` operator will throw a `TypeError`. Consider guarding this with `if (!raw || typeof raw !== 'object') return raw` to fail gracefully down the line instead of crashing.

src/lib/config.ts
```diff
@@ -153,11 +189,17 @@ export function validateConfigForDoctor(config: Record<string, unknown>): string
...
-            !UPDATE_CHANNELS.has(config.updateChannel as UpdateChannel))
+        config.update_channel !== undefined &&
+        (typeof config.update_channel !== 'string' ||
+            !UPDATE_CHANNELS.has(config.update_channel as UpdateChannel))
     ) {
-        issues.push('updateChannel must be one of: stable, pre-release')
+        issues.push('update_channel must be one of: stable, pre-release')
+    }
+
+    if (config.updateChannel !== undefined) {
```
doistbot: [P2] This new legacy-key branch only emits the migration notice; it no longer validates the legacy value itself. For example, `validateConfigForDoctor({ updateChannel: 'beta' })` now misses the invalid-channel warning, even though `migrateLegacyKeys()` will later carry that raw value forward into `update_channel`. Please apply the same `stable` / `pre-release` validation to `updateChannel` before adding the legacy-key warning.

src/lib/config.ts
```diff
@@ -50,18 +53,41 @@ export interface Config {
     // Auth metadata persisted alongside the token to track OAuth scope.
     authMode?: AuthMode
     authScope?: string
-    updateChannel?: UpdateChannel
+    // Snake_case to match cli-core's update command, which reads this key
```
doistbot: [P2] Exposing `update_channel` on the exported `Config` type leaks a disk-format/cli-core implementation detail through the whole app. `lib/config.ts` is already the persistence seam, so commands and other library code can keep the logical camelCase field and translate to/from `update_channel` inside the read/write helpers. Leaving the public API in snake_case creates a one-off exception every future config caller has to remember.

src/lib/config.ts
```diff
@@ -109,9 +138,16 @@ export async function setConfig(config: Config): Promise<void> {
...
+/**
+ * Read-merge-write wrapper around cli-core's `updateConfig`. Inlines the
+ * read+write rather than calling cli-core's atomic variant so the on-disk
+ * file is rewritten in the canonical snake_case shape (the legacy
+ * `updateChannel` field is dropped here rather than lingering across update
+ * calls). Race-free in practice — the config file is single-user.
+ */
 export async function updateConfig(updates: Partial<Config>): Promise<void> {
-    await updateConfigCore<Config>(getConfigPath(), updates)
+    const current = await getConfig()
```
doistbot: [P2] This changes `updateConfig()` from a shared partial-write abstraction into a plain read/merge/write. That reintroduces lost-update races between concurrent `tw` processes: two callers can read the same file, apply unrelated updates, and the last write wins. Since this helper is used across auth/workspace flows, it should preserve cli-core's atomic update semantics and handle legacy-key cleanup separately.

src/commands/doctor.test.ts
```diff
@@ -161,7 +161,7 @@ describe('doctor command', () => {
                 pendingSecureStoreClear: 'yes',
                 currentWorkspace: 'abc',
                 authMode: 'admin',
-                updateChannel: 'beta',
+                update_channel: 'beta',
```
doistbot, line 164: [P2] This suite now covers canonical `update_channel` warnings, but it still doesn't exercise the new `tw doctor` path for a legacy on-disk `{ "updateChannel": ... }` config. Please add a command-level test here that feeds `mockReadFile` a legacy key and asserts the emitted warning includes the migration message while not reporting `updateChannel` as an unrecognized key.

src/lib/config.ts
```diff
@@ -109,9 +138,16 @@ export async function setConfig(config: Config): Promise<void> {
     await writeConfigCore(getConfigPath(), config)
 }
 
-/** Thin wrapper around cli-core's `updateConfig`. */
+/**
+ * Read-merge-write wrapper around cli-core's `updateConfig`. Inlines the
+ * read+write rather than calling cli-core's atomic variant so the on-disk
+ * file is rewritten in the canonical snake_case shape (the legacy
+ * `updateChannel` field is dropped here rather than lingering across update
```
doistbot: [P2] Dropping the legacy `updateChannel` key on write makes this a one-step rename of persisted data. Our version-overlap guidance says field renames need a multi-step migration; older released twist-cli builds still reading `updateChannel` will silently lose the configured channel after any config write. Please keep dual-write compatibility for at least one release (or translate at the cli-core boundary) before removing the legacy key from disk. https://handbook.doist.com/doc/standard-service-production-readiness-qgNRng8TJx

doistbot (commented):
This PR updates the configuration schema to align the `updateChannel` key with `@doist/cli-core`'s snake_case requirement and introduces migration handling for legacy configs. While this effectively prepares the CLI for delegating update commands and includes helpful diagnostic warnings, there are a few areas to refine regarding persistence and backward compatibility. Specifically, there are opportunities to maintain atomic writes to avoid race conditions, type-guard manual edits, validate legacy values during doctor checks, keep the public API in camelCase, expand test coverage, and implement a dual-write phase to support older CLI versions.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F211&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A854baef0-4cb4-11f1-9d0d-6b77c3e6b35d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A854baef0-4cb4-11f1-9d0d-6b77c3e6b35d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/config.ts
scottlovegrove: Fixed in c74d15c — `fromDiskShape` (the replacement for `migrateLegacyKeys`) now guards with `if (!raw || typeof raw !== "object" || Array.isArray(raw)) return {}`. A manually-edited config with a top-level `null`, primitive, or array now resolves to an empty config object instead of crashing on the `in` check. Test added: `getConfig guards against a manually-edited config that is not an object`.

src/lib/config.ts
scottlovegrove: Fixed in c74d15c — `validateConfigForDoctor` now applies the same `stable | pre-release` check to both keys. `validateConfigForDoctor({ updateChannel: "beta" })` emits `"updateChannel must be one of: stable, pre-release"`; same shape check for the canonical `update_channel`. Added command-level doctor test that feeds `mockReadFile` a legacy `{ "updateChannel": "beta" }` and asserts the warning surfaces.

src/lib/config.ts
scottlovegrove: Fixed in c74d15c — agreed, this was the right call. `Config.updateChannel` stays camelCase (the in-memory shape every caller sees); `update_channel` is now purely a persistence concern handled by `fromDiskShape` / `toDiskShape` at the read/write seam. Reverted all the downstream sites (`src/lib/update.ts`, `src/commands/update/switch.ts`, `src/commands/config/view.ts`) back to `config.updateChannel`. The disk-format detail no longer leaks past `src/lib/config.ts`.

src/lib/config.ts
scottlovegrove: Fixed in c74d15c — `updateConfig` is back to delegating to cli-core's `updateConfigOrThrow` for atomicity. The dual-write translation runs on the partial before it reaches cli-core, so the atomic merge is preserved. Test `updateConfig delegates to cli-core (atomic) with dual-write translation` asserts cli-core's function is the one called.

src/commands/doctor.test.ts
scottlovegrove: Fixed in c74d15c — added two command-level doctor tests in `src/commands/doctor.test.ts`:

- `reads legacy updateChannel from on-disk config and reports the channel` feeds `mockReadFile` `{ updateChannel: "pre-release" }`, asserts doctor reports `PASS CLI is up to date on pre-release` and that the legacy key is not flagged as `unrecognized`.
- `flags invalid legacy updateChannel value in the validator` feeds `{ updateChannel: "beta" }` and asserts the validator warning surfaces in doctor output.

src/lib/config.ts
scottlovegrove: Fixed in c74d15c — `toDiskShape` now dual-writes both `updateChannel` and `update_channel` to disk when the channel is set. Older twist builds keep reading the camelCase key through the overlap window; cli-core's update command (PR2) will read the snake_case one. The camelCase write is intended to be dropped in a release or two once all deployed twist versions read `update_channel`. Test `setConfig dual-writes the channel field (camelCase + snake_case)` locks the behavior down.

One caveat worth flagging: cli-core's own update command (which lands in the follow-up PR) writes only `update_channel` via its internal atomic-update — it will not dual-write. So once that PR ships and a user runs `tw update switch`, the on-disk `updateChannel` value goes stale relative to `update_channel`. Older twist still reads the stale camelCase value. The window where this matters is narrow (a user has to be actively running two twist versions across a config-write), and the failure mode is graceful (channel preference resets to "stable" effectively), but it's worth noting that strict version-overlap is bounded by what cli-core writes, not what twist writes.

pawelgrimm (commented):
🚀

PR #215, opened 2026-05-11
refactor: delegate auth flow to @doist/cli-core/auth

scottlovegrove:
## Summary

- Drops the local OAuth coordination (`src/lib/pkce.ts`, `oauth.ts`, `oauth-server.ts`, `commands/auth/login.ts` body) and delegates the flow to `@doist/cli-core/auth`'s `runOAuthFlow` + `attachLoginCommand`.
- New `TwistAuthProvider` (`src/lib/auth-provider.ts`) owns the twist-specific behaviour:
  - `prepare()` — Dynamic Client Registration via `POST /oauth/register`.
  - `authorize()` — builds the twist authorize URL using cli-core's `generateVerifier` / `deriveChallenge`.
  - `exchangeCode()` — `client_secret_basic` token exchange with the PKCE verifier.
  - `validateToken()` — `users.getSessionUser` to populate the `TwistAccount`.
- New `TwistTokenStore` adapter wraps the existing `saveApiToken` / `probeApiToken` / `clearApiToken` so cli-core writes the token through the same dual-storage (keyring + config fallback) twist already uses elsewhere.
- Branded success / error HTML (logo, terminal mockup, 30-second auto-close) preserved byte-for-byte in `src/lib/auth-pages.ts` and passed to `attachLoginCommand` via `renderSuccess` / `renderError`.
- `tw auth login` gains `--callback-port`, `--json`, and `--ndjson` for free from `attachLoginCommand` (was `--read-only` only).
- Net: 9 files changed, +618 / −764. 580/580 tests pass. Login-flow tests retargeted to provider + store units in `src/lib/auth-provider.test.ts`.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check` (oxlint + oxfmt)
- [x] `npm test` (580 tests pass)
- [x] `npm run build`
- [x] Smoke: `tw auth logout && tw auth login` against twist.com — branded success page renders, token persisted, green checkmark printed
- [x] `tw auth login --help` lists `--read-only`, `--callback-port`, `--json`, `--ndjson`

PR #216, opened 2026-05-11
feat(auth): support multiple authenticated Twist accounts (v2 on top of cli-core)

scottlovegrove:
## Summary

Re-implementation of #206 on top of \`main\`, which has since switched the OAuth flow to \`@doist/cli-core/auth\` (#215). A mechanical rebase wasn't possible — \`oauth-server.ts\` was deleted in #215 and \`auth.ts\` was rewritten — so this is a fresh PR closing #206.

Replaces single-account auth state with a per-user store keyed by Twist user id. The active account is selected via the new global \`--user <id|email>\` flag (parsed once and stripped from argv so every command honours it without per-command wiring) or via \`user.default_user\` in config; with one stored account that account is implicit.

### Storage

- v2 schema uses snake_case (\`config_version\`, \`user.default_user\`, \`users[]\`, \`users[].auth_mode\`, \`users[].api_token\`, \`users[].current_workspace\`, …) to mirror todoist-cli so a future cli-core extraction of the multi-user shape is a drop-in.
- Tokens move into per-user keyring slots \`user-<id>\` via \`accountForUser(id)\`; the legacy \`api-token\` slot is consulted only via the v1 fallback path and during migration.
- A best-effort postinstall migration (\`src/lib/migrate-auth.ts\`) upgrades v1 installs by hitting \`/api/v3/users/get_session_user\`. Failure leaves v1 state untouched and the runtime resolver falls back to the legacy token until the next attempt.

### Resolver (\`src/lib/auth.ts::resolveActiveUser\`)

- env > \`--user\` > default > single > legacy fallback.
- Legacy fallback is gated on \`!Array.isArray(config.users)\` so a logged-out v2 install (\`users: []\`) can't silently re-auth from a stale keyring entry.
- \`upsertUser\` rolls back the keyring write if config write fails for a *new* user; existing users keep the working secret.
- \`removeUserById\` writes config first then deletes the secret — partial failures leave the account functional and retry is safe.

### Commands

- \`tw account list\` / \`tw account use <ref>\` — inspect / set default. (Named \`account\` rather than \`user\` because \`tw user\` already shows current Twist user info.)
- \`tw auth login\` (via cli-core flow's \`TokenStore.set\` adapter in auth-provider.ts) writes per-user and identifies the user via \`validateToken\`.
- \`tw auth token\` probes \`users.getSessionUser\` to identify the account before upserting.
- \`tw auth logout\` honours \`--user\` to target a specific account.
- \`tw auth status\` shows the active account and lists the others.

### Workspace per-account (\`api.ts\`)

- \`config.currentWorkspace\` (top-level, v1) is migrated lazily onto the default/only user on the first workspace lookup. After that, every account carries its own \`current_workspace\`.
- \`inbox.ts\` and \`channel/threads.ts\` route batch responses through the existing \`assertBatchData\`, so a future API error surfaces as \`BATCH_FAILED\` rather than a \`TypeError\` on \`.map\`.

### \`--user\` flag validation (\`global-args.ts::validateUserFlag\`)

- Pre-Commander rejection of bare \`--user\`, \`--user=\` (empty), and \`--user <known-subcommand>\` (typo: forgot the value).
- \`--\` terminator honoured: \`--user\`-shaped tokens after \`--\` are positional content.
- \`lib/auth\` accepts \`opts.ref\` rather than reading from global-args, keeping the lib/CLI layer boundary clean.

### Tests

- New: \`users.test.ts\`, \`auth.test.ts\` (multi-user resolver + upsert/clear/remove/setDefault), \`migrate-auth.test.ts\`, \`global-args.user.test.ts\` (\`--user\` parsing + validation + strip).
- Updated: \`inbox.test.ts\` and \`channel/threads.test.ts\` mocks include \`assertBatchData\`; \`postinstall.test.ts\` asserts the migration is invoked.

### Notes

- The \`~60s hang fix\` from #206's \`cbc9678\` is already in cli-core (\`flow.js\` calls \`closeAllConnections()\`), so it was naturally dropped.
- Original \`auth-provider.test.ts\` and \`auth/auth.test.ts\` were dropped — the new TokenStore behaviour is exercised indirectly via \`auth.test.ts\`'s upsert tests. A dedicated integration test for \`attachLoginCommand\` + the multi-user store would be a useful follow-up but isn't blocking.

## Test plan

- [x] \`npm run type-check\`
- [x] \`npm run lint:check\` (oxlint + oxfmt)
- [x] \`npm test\` — 585 pass
- [x] \`npm run build\` + \`npm run sync:skill\`
- [x] \`tw --help\` lists \`--user <id|email>\`
- [x] \`tw account --help\` and \`tw account list --help\` work
- [x] \`tw --user inbox\` → \`USER_FLAG_INVALID\` with hint
- [x] \`tw --user\` → \`USER_FLAG_INVALID\` with hint
- [ ] Live OAuth round-trip against twist.com (manual)
- [ ] v1 → v2 migration smoke against a real legacy install (manual)

Closes #206.


scottlovegrove:
@rfgamaral Don't worry about this PR just yet.

scottlovegrove:
Closing — splitting into 4 stacked PRs against `scottl/multi-user-base` so each chunk is reviewable independently. Single semantic-release fires when the integration branch lands in `main`. See follow-up PRs.

doistbot (commented):
This PR successfully transitions the CLI to a multi-account authentication architecture built on top of `@doist/cli-core`, introducing per-user storage, token migration, and a global user selection flag. The changes provide a significant structural improvement that nicely aligns our tooling and prepares the ground for smoother multi-tenant workflows. To fully polish the implementation, a few adjustments are needed to properly thread the user flag through auth metadata, preserve workspace states during upserts and legacy fallbacks, reduce redundant keychain reads, add machine-readable output to the new commands, and restore essential test coverage for the updated auth flows.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F216&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A45151ad0-4d66-11f1-908f-89ac7de5e9ed%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A45151ad0-4d66-11f1-908f-89ac7de5e9ed&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-        return {
-            token: envToken,
-            metadata: { authMode: 'unknown', source: 'env' },
+    const resolved = await resolveActiveUser()
+    return { token: resolved.token, metadata: resolvedToMetadata(resolved) }
+}
+
+export async function getAuthMetadata(): Promise<AuthMetadata> {
+    try {
+        const resolved = await resolveActiveUser()
```
doistbot, line 198: [P1] `getAuthMetadata()` resolves the account without the selected `--user` ref. `ensureWriteAllowed()` and `tw auth status` both depend on this, so `tw --user other@example.com ...` can report or enforce the default account's auth mode instead of the requested account's. Thread the requested ref through here (or pass it from callers) so metadata comes from the same account as `getTwistClient()`.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
+        if (!(error instanceof SecureStoreUnavailableError)) throw error
     }
 
-    const secureStore = createSecureStore()
+    const userRecord: StoredUser = storedSecurely
+        ? baseRecord
+        : { ...baseRecord, api_token: trimmedToken }
+
+    let next = ensureV2Shape(config)
+    next = upsertStoredUser(next, userRecord).config
```
doistbot, line 264: [P1] This wholesale upsert replaces any existing stored user record with `userRecord`, which drops local-only fields such as `current_workspace`. Re-authenticating an existing account via `tw auth login` or `tw auth token` will reset that account's workspace selection. Merge onto the existing record instead of overwriting it.

src/index.ts
```diff
@@ -199,21 +203,45 @@ if (process.argv[2] === 'completion-server') {
...
+// clear CliError with hints instead of Commander's terser usage message.
+{
+    const knownCommands = new Set<string>([
+        ...Object.keys(commands),
+        ...Object.keys(commandAliases),
+    ])
+    const userArgs = process.argv.slice(2)
+    const validation = validateUserFlag(userArgs, knownCommands)
+    if (!validation.ok) {
+        const err = new CliError('USER_FLAG_INVALID', validation.message, validation.hints, 'info')
```
doistbot, line 217: [P2] The invalid `--user` branch sets `process.exitCode` and skips `program.parseAsync()`, but the early spinner may already have been started above when a command was detected. For inputs like `tw --user inbox`, this leaves the spinner running because `stopEarlySpinner()` is only reached in the parse path. Stop it before returning from this branch.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
     }
 
-    const config = await getConfig()
-    if (config.currentWorkspace) {
-        return config.currentWorkspace
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
+
+    // Env-token / legacy-fallback flows have no persistent slot to write
+    // into; resolve the workspace freshly from the API each call.
+    if (!resolved.id) {
```
doistbot, line 273: [P2] This branch handles both env tokens and the v1 runtime fallback, but it no longer consults legacy `config.currentWorkspace`. If postinstall migration hasn't succeeded yet, a previously selected workspace is ignored and commands fall back to the API default/first workspace instead. Preserve the v1 workspace here (or migrate it before branching) so legacy installs keep their current selection until migration completes.

src/commands/auth/token.ts
```diff
@@ -44,7 +45,30 @@ export async function loginWithToken(token?: string): Promise<void> {
...
     }
-    const saveResult = await saveApiToken(token.trim(), { authMode: 'unknown' })
-    console.log(chalk.green('✓'), 'API token saved successfully!')
-    logTokenStorageResult(saveResult, 'Token stored securely in the system credential manager')
+    const trimmed = token.trim()
+
+    // Identify the user behind the token so it lands in the right slot.
+    let user: { id: number; name: string; email: string }
+    try {
+        const probe = createWrappedTwistClient(trimmed)
```
doistbot, line 53: [P2] This manual-token probe reimplements the same "identify account from token" call path that already exists in `src/lib/auth-provider.ts` (`createWrappedTwistClient(token).users.getSessionUser()`). Please extract that lookup into a shared helper, or reuse the provider validator here, so OAuth login and `tw auth token` can't drift in how they resolve accounts and surface token-validation errors.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-    if (envToken) {
-        return { authMode: 'unknown', source: 'env' }
-    }
+/**
+ * Set the default user, looked up by id or email. Returns the resolved
+ * user so callers can show "switched to <email>" without re-resolving.
+ */
+export async function setDefaultUserId(ref: string): Promise<StoredUser> {
+    const config = ensureV2Shape(await getConfig())
+    const found = findUserByRef(config, ref)
```
doistbot, line 369: [P2] `src/lib/users.ts` already exposes `requireUserByRef()` for this exact find-or-throw flow, and `resolveTargetUser()` duplicates the same pattern a few lines above. Reusing that helper here would keep account lookup and `ACCOUNT_NOT_FOUND` behavior centralized instead of maintaining multiple hand-written branches.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
+    const { currentWorkspace: _drop, ...rest } = config
+    return { ...rest, users: next }
+}
+
+/**
+ * Persist the active user's workspace selection. Used by `tw workspace use`.
+ */
+export async function setActiveAccountWorkspace(workspaceId: number): Promise<void> {
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
+    if (!resolved.id) return
```
doistbot, line 353: [P2] Returning silently here makes `setActiveAccountWorkspace()` an unsupported-mode no-op for `TWIST_API_TOKEN` / legacy auth, while `tw workspace use` still reports success. That leaves a brittle API boundary where callers cannot tell whether persistence actually happened. Surface an explicit error/result instead of silently skipping the write.

src/lib/auth-provider.ts
```diff
@@ -278,16 +282,17 @@ export function createTwistTokenStore(): TwistTokenStore {
...
+                    // install — no v2 identity to round-trip.
                     return null
                 }
                 return {
                     token,
                     account: {
-                        id: String(metadata.authUserId),
-                        label: metadata.authUserName,
+                        id: metadata.accountId,
+                        label: metadata.email,
```
doistbot, line 294: [P2] `TwistAccount.label` is documented and produced by `validateToken()` as the display name, but `active()` reconstructs it from `metadata.email`. That makes the token-store adapter non-round-trippable and gives cli-core two different identity shapes depending on the code path. Preserve the stored name in metadata (or read it from config here) so `active()` returns the same semantic account shape that `set()`/`validateToken()` use.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
         return flagValue
     }
 
-    const config = await getConfig()
-    if (config.currentWorkspace) {
-        return config.currentWorkspace
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
```
doistbot, line 269: [P2] `resolveActiveUser` reads the user token from the OS credential manager (`secureStore.getSecret()`), which can be slow. Since `getTwistClient()` and `getCurrentWorkspaceId()` are typically both called during a single command execution, this introduces a redundant keychain access on the critical path. Consider caching the `resolveActiveUser()` result in memory for the duration of the process, or splitting out a helper that resolves the user identity without fetching the token.

src/commands/inbox.test.ts
```diff
@@ -10,6 +10,12 @@ const apiMocks = vi.hoisted(() => ({
 vi.mock('../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
```
doistbot, line 13: [P2] Re-implementing `assertBatchData` inside the test mock duplicates the real function's logic and makes the test brittle if the implementation changes (e.g., throwing a specific `CliError`). Use `importOriginal` to retain the actual implementation of this utility.

```typescript
vi.mock('../lib/api.js', async (importOriginal) => {
    const actual = await importOriginal<typeof import('../lib/api.js')>()
    return {
        ...actual,
        getTwistClient: apiMocks.getTwistClient,
        getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
    }
})
```

src/lib/auth.test.ts
```diff
@@ -1,273 +1,220 @@
...
-        mocks.getConfig.mockReset()
-        mocks.setConfig.mockReset()
-        mocks.updateConfig.mockReset()
-        mocks.unlink.mockReset()
-        mocks.secureTokenStore.getSecret.mockReset()
-        mocks.secureTokenStore.setSecret.mockReset()
-        mocks.secureTokenStore.deleteSecret.mockReset()
         delete process.env.TWIST_API_TOKEN
+        setConfigState({})
+        mockCreateSecureStore.mockImplementation(() => memorySecureStore())
```
doistbot, line 64: [P3] Consider using a `Map`-backed stateful mock here (similar to the one in `uses --user <ref> over the default`) instead of returning a new instance on every call. Currently, operations like `upsertUser` and `clearApiToken` interact with completely disjoint store instances when called in the same test. While it passes now, it creates an unrealistic mock environment that could hide bugs if assertions become stricter.

src/commands/auth/auth.test.ts
```diff
@@ -1,372 +0,0 @@
-import { Command } from 'commander'
```
doistbot, line 1: [P2] Removing this suite leaves no command-level coverage for the auth flows changed in this PR. The new `src/lib/auth.test.ts` only exercises storage helpers, so regressions in `auth token`'s probe/upsert flow, `auth status` output, `logout --user`, or the cli-core login wiring would now go uncaught. Please keep a thin `registerAuthCommand` integration suite.

src/lib/auth-provider.test.ts
```diff
@@ -1,232 +0,0 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
```
doistbot, line 1: [P2] Deleting the dedicated cli-core adapter tests leaves `createTwistAuthProvider()` / `createTwistTokenStore()` untested. The replacement auth tests never instantiate these units, so `active()` round-tripping account metadata, `set()`/`clear()`, and `validateToken()` no longer have direct coverage.

src/lib/auth.test.ts
```diff
@@ -1,273 +1,220 @@
...
-
-        await expect(clearApiToken()).resolves.toEqual({
-            storage: 'config-file',
-            warning:
-                'system credential manager unavailable; local auth state cleared in /home/user/.config/twist-cli/config.json',
-        })
-        expect(mocks.setConfig).toHaveBeenCalledWith({
-            currentWorkspace: 7,
-            pendingSecureStoreClear: true,
+    it('upsertUser persists a v2 user record and auto-defaults the first one', async () => {
```
doistbot, line 171: [P2] This new suite only asserts happy paths. The most failure-prone behavior added in `auth.ts` is the rollback/partial-failure logic (`upsertUser()` undoing a brand-new keyring write when `setConfig` fails, and `removeUserById()` tolerating a keyring delete failure after config is updated), but none of those branches are covered here.

src/commands/channel/threads.test.ts
```diff
@@ -14,6 +14,12 @@ const refsMocks = vi.hoisted(() => ({
 vi.mock('../../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
```
doistbot, line 17: [P2] This mock update keeps the existing happy-path tests compiling, but it still doesn't exercise the new behavior the PR calls out: batch failures should surface as `BATCH_FAILED` instead of blowing up on `.map`. Please add a case with `{ code: 500, data: null }` and assert the command rejects with the structured CLI error (the same gap exists in `inbox.test.ts`).

src/commands/account/index.ts
```diff
@@ -0,0 +1,30 @@
...
+        .description('Manage stored Twist accounts (multi-user)')
+
+    account
+        .command('list')
+        .description('List all stored Twist accounts')
+        .option('--json', 'Output as JSON')
+        .action(listAccountsCommand)
+
+    account
+        .command('use <ref>')
```
doistbot, line 17: [P2] `account use` is a new state-changing command, but it only emits human text. The repo’s command-output convention is to provide `--json` for mutating commands when there’s scripting value; without it, agents/scripts can’t switch the default account reliably. Please add a `--json` option and have `useAccountCommand` return a minimal machine-readable object via `formatJson()`.

PR #217, opened 2026-05-11
refactor(auth): [1/4] multi-user storage foundation

scottlovegrove:
**Part 1 of 4 stacked PRs** that re-implement #206 on top of the cli-core auth integration (#215). All four target the integration branch `scottl/multi-user-base` so semantic-release doesn't fire per PR — a single release happens when the integration branch lands in `main`.

## Order

- **#217 this PR** — refactor: foundation (storage + resolver, no user-facing surface)
- #219 — fix: per-account `current_workspace`
- #220 — feat: postinstall v1 → v2 migration

## Summary

Storage + resolver primitives without any user-facing surface. Single-user installs keep working via the v1 legacy fallback in `resolveActiveUser`.

- **`config.ts`** — v2 schema with snake_case fields (`config_version`, `user.default_user`, `users[]`, `users[].auth_mode`, `users[].api_token`, `users[].current_workspace`, …) mirroring todoist-cli so a future cli-core extraction is a drop-in. Legacy v1 keys kept on the type for migration.
- **`secure-store.ts`** — new `accountForUser(id)` returns `user-<id>`. `createSecureStore` takes an account-name parameter (defaults to `LEGACY_ACCOUNT_NAME` so legacy reads still work).
- **`users.ts`** (NEW) — pure data layer on `config.users[]`: `getStoredUsers`, `findUserByRef`, `upsertStoredUser`, `removeStoredUser`, `setDefaultUser`, `updateStoredUser`. `AccountNotFoundError` / `NoAccountSelectedError` thrown from the resolver (codes are `ACCOUNT_*` because the user-facing command will be `tw account` — `tw user` already shows session info).
- **`auth.ts`** (rewritten) — `resolveActiveUser({ ref })` returns the active user's token + identity. Env > `--user` > default > single > legacy fallback. Legacy fallback is gated on `!Array.isArray(config.users)` so a logged-out v2 install (`users: []`) can't silently re-auth from a stale `api-token` keyring entry. `upsertUser` rolls back the keyring write if config write fails for a *new* user; existing users keep the working secret. `removeUserById` writes config first then deletes the secret — partial failures leave the account functional and retry is safe. Backwards-compatible `getApiToken` / `probeApiToken` / `getAuthMetadata` still work.
- **`auth-provider.ts`** — `TwistTokenStore.set` delegates to `upsertUser` so cli-core's `runOAuthFlow` lands tokens in the per-user slot. `TwistAccount` carries `email` (needed by `upsertUser`).
- **`commands/auth/token.ts`** — probes `users.getSessionUser` with the provided token to identify the user, then calls `upsertUser`. Wraps probe failures as `INVALID_TOKEN` with hints.
- **`errors.ts`** — adds `ACCOUNT_NOT_FOUND`, `NO_ACCOUNT_SELECTED`, `INVALID_ACCOUNT`, `CONFIG_WRITE_FAILED`, `USER_FLAG_INVALID`, `MISSING_REF`.

The old `auth.test.ts` / `auth-provider.test.ts` / `auth/auth.test.ts` were tied to the dropped `saveApiToken` single-user API and have been replaced with new `users.test.ts` + multi-user `auth.test.ts` exercising the data layer + resolver + upsert/clear/remove paths.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 562 pass


src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-                authUserName: config.authUserName,
-                source: 'config-file',
-            },
-        }
+    const previouslyExisted = getStoredUsers(config).some((u) => u.id === input.id)
+    // Always default the first stored user, even if `user.default_user`
+    // points at an orphaned id, so a stale pointer can't wedge resolution.
+    const shouldSetDefault = getStoredUsers(config).length === 0
+
+    const baseRecord: StoredUser = {
```
doistbot, line 242: [P2] `upsertUser` completely replaces the user record but drops the existing `current_workspace`. If a user re-authenticates (e.g. via `tw auth login` or `tw auth token`), they will lose their active workspace selection. Consider looking up the existing user and carrying over `current_workspace` into `baseRecord`.

src/lib/auth-provider.ts
```diff
@@ -278,16 +282,17 @@ export function createTwistTokenStore(): TwistTokenStore {
...
+                    // install — no v2 identity to round-trip.
                     return null
                 }
                 return {
                     token,
                     account: {
-                        id: String(metadata.authUserId),
-                        label: metadata.authUserName,
+                        id: metadata.accountId,
+                        label: metadata.email,
```
doistbot, line 294: [P2] `TwistTokenStore.active()` uses `metadata.email` for the account `label`, whereas `validateToken` uses `user.name`. This causes the display label in `cli-core` prompts to fluctuate between the user's name and email across sessions. Consider adding `name` to `AuthMetadata` (populated in `resolvedToMetadata`) and using `label: metadata.name ?? metadata.email` here so the account name consistently round-trips.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-
-function withoutAuthFallbackState(config: Config): Config {
-    const { token: _token, pendingSecureStoreClear: _pending, ...rest } = config
+function stripLegacyAuthFields(config: Config): Config {
+    const {
+        token: _t,
+        authMode: _m,
+        authScope: _s,
+        pendingSecureStoreClear: _p,
+        currentWorkspace: _w,
```
doistbot, line 512: [P1] `stripLegacyAuthFields()` now drops top-level `currentWorkspace` on every auth write, but `src/lib/api.ts` still reads `config.currentWorkspace` as the active workspace. That means a successful `tw auth login` / `tw auth token` for an existing install silently resets the user's workspace selection, and later commands can run against a different workspace than before. Please preserve `currentWorkspace` until the per-account migration lands, or migrate it onto the selected user before removing the legacy field.

src/lib/users.ts
```diff
@@ -0,0 +1,134 @@
...
+    } else {
+        users.push(next)
+    }
+    return { config: { ...config, users }, replaced }
+}
+
+export function removeStoredUser(config: Config, id: string): Config {
+    const users = getStoredUsers(config).filter((u) => u.id !== id)
+    const next: Config = { ...config, users }
+    if (next.user?.default_user === id) {
```
doistbot, line 113: [P2] Clearing `default_user` outright when the default account is removed leaves 2+ remaining accounts in a state that this PR cannot recover from: `resolveActiveUser()` starts throwing `NO_ACCOUNT_SELECTED`, but the `--user` flag / `tw account use` commands are not shipped yet. A user can hit this today by storing 3 accounts and running `tw auth logout` once. Please promote one remaining account to the new default (or keep the old behavior gated until selection commands exist).

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-        const storedToken = await secureStore.getSecret()
-        if (storedToken?.trim()) {
-            return storedToken
-        }
-    } catch (error) {
-        if (!(error instanceof SecureStoreUnavailableError)) {
-            throw error
-        }
+    if (requestedRef) {
+        const found = findUserByRef(config, requestedRef)
```
doistbot, line 163: [P2] This `findUserByRef` + throw pattern duplicates the logic of the `requireUserByRef` helper that was newly added and exported in `src/lib/users.ts`. Consider importing and using `requireUserByRef(config, requestedRef)` here, as well as below in `setDefaultUserId` (line 369).

src/commands/auth/token.ts
```diff
@@ -44,7 +45,30 @@ export async function loginWithToken(token?: string): Promise<void> {
...
-    logTokenStorageResult(saveResult, 'Token stored securely in the system credential manager')
+    const trimmed = token.trim()
+
+    // Identify the user behind the token so it lands in the right slot.
+    let user: { id: number; name: string; email: string }
+    try {
+        const probe = createWrappedTwistClient(trimmed)
+        user = await probe.users.getSessionUser()
+    } catch (error) {
+        const detail = error instanceof Error && error.message ? `: ${error.message}` : ''
```
doistbot, line 56: [P3] The logic to extract and format the error message (`error instanceof Error && error.message ? ...`) is duplicated here and multiple times in the new `src/lib/auth.ts` additions (lines 290 and 343). Consider extracting a shared helper (e.g., `getErrorMessage(error)`) to standardize this defensive error parsing across the module.

doistbot (commented):
This PR lays a solid foundation for multi-user authentication by introducing the v2 configuration schema and integrating with the cli-core auth primitives. The structural shift cleanly prepares the groundwork for upcoming account management features while maintaining legacy fallback paths. However, there are a few important areas to revisit, particularly restoring deleted integration tests for OAuth and CLI commands, preventing the premature removal of the legacy active workspace field, and addressing several edge cases around state preservation, keyring I/O optimization, error handling consistency, and aligning hints and skill content with currently shipped capabilities.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F217&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab7d250a0-4d67-11f1-9945-5659914f3bae%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab7d250a0-4d67-11f1-9945-5659914f3bae&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-    }
+/**
+ * Set the default user, looked up by id or email. Returns the resolved
+ * user so callers can show "switched to <email>" without re-resolving.
+ */
+export async function setDefaultUserId(ref: string): Promise<StoredUser> {
+    const config = ensureV2Shape(await getConfig())
+    const found = findUserByRef(config, ref)
+    if (!found) throw new AccountNotFoundError(ref)
+    await setConfig(stripLegacyAuthFields(setDefaultUserInConfig(config, found.user.id)))
```
doistbot, line 371: [P2] `setDefaultUserId()` writes config directly instead of reusing the CONFIG_WRITE_FAILED wrapping pattern introduced a few lines above in `upsertUser()` and `removeUserById()`. That means `tw account use` will leak raw fs/permission errors while the other account mutations return structured CLI errors. Please route this through the same shared config-write helper so all account writes behave consistently.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-        return {
-            token: envToken,
-            metadata: { authMode: 'unknown', source: 'env' },
+    const resolved = await resolveActiveUser()
+    return { token: resolved.token, metadata: resolvedToMetadata(resolved) }
+}
+
+export async function getAuthMetadata(): Promise<AuthMetadata> {
+    try {
+        const resolved = await resolveActiveUser()
```
doistbot, line 198: [P2] `getAuthMetadata()` now goes through `resolveActiveUser()`, which loads the active token from the secure store before returning metadata. `ensureWriteAllowed()` calls this before every mutating API call, so each write now pays an extra keyring lookup even though `auth_mode`/`auth_scope` already live in config. A metadata-only resolver that selects the account from config without fetching the secret would avoid this added latency.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
+    try {
+        await setConfig(next)
+    } catch (error) {
+        const detail = error instanceof Error && error.message ? `: ${error.message}` : ''
+        throw new CliError('CONFIG_WRITE_FAILED', `Could not update ${getConfigPath()}${detail}`, [
+            'Check file permissions on ~/.config/twist-cli/, then re-run the command',
+        ])
+    }
+
+    const secureStore = createSecureStore(accountForUser(id))
```
doistbot, line 349: [P2] `removeUserById()` always opens the per-user keyring entry and calls `deleteSecret()`. For users stored via the plaintext `api_token` fallback, there is no secret to delete, so this is unnecessary keyring I/O on every logout/remove of that account. Keeping the removed record (or at least whether it had `api_token`) before pruning would let this path skip the extra keyring round-trip.

src/commands/auth/auth.test.ts
```diff
@@ -1,372 +0,0 @@
-import { Command } from 'commander'
```
doistbot, line 1: [P1] Deleting this file drops all integration tests for the `tw auth` CLI commands (interactive prompts, `--json` formatting, `logout` behavior, and `login` wiring). While the underlying single-user token API was removed, the commands themselves still exist and were updated to use `upsertUser`. These integration tests should be updated to mock the new functions instead of being deleted. (When restoring, please also add a test case for the new `getSessionUser()` probe failure path in `tw auth token`).

src/lib/auth-provider.test.ts
```diff
@@ -1,232 +0,0 @@
-import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
```
doistbot, line 1: [P1] The `createTwistAuthProvider` tests (covering OAuth handshakes, PKCE verifiers, and token exchange) are completely independent of the dropped `saveApiToken` API. Deleting this file leaves the twist-specific OAuth integration logic untested. Please restore the `createTwistAuthProvider` test suites, and update the `createTwistTokenStore` tests to verify the new multi-user integration.

src/lib/auth.test.ts
```diff
@@ -1,273 +1,220 @@
...
-            token: 'legacy_token_123456',
-            currentWorkspace: 7,
-        })
-
-        const { clearApiToken } = await import('./auth.js')
-
-        await expect(clearApiToken()).resolves.toEqual({ storage: 'secure-store' })
-        expect(mocks.secureTokenStore.deleteSecret).toHaveBeenCalled()
-        expect(mocks.setConfig).toHaveBeenCalledWith({ currentWorkspace: 7 })
+describe('upsertUser + clearApiToken + removeUserById + setDefaultUserId + listStoredUsers', () => {
```
doistbot, line 164: [P2] This new suite only covers happy paths. The most failure-prone behavior introduced in this refactor is the rollback/partial-failure logic (`upsertUser()` deleting a newly written secret when `setConfig()` fails, and `removeUserById()`/`clearApiToken()` tolerating keyring-delete failures after the config write), but none of those branches are tested here. Add targeted tests that make `setConfig` or `deleteSecret` throw.

src/lib/users.ts
```diff
@@ -0,0 +1,134 @@
...
+ * command surface — twist already exposes `tw user` for displaying
+ * Twist API user info, so the multi-identity command lives under
+ * `account`. The on-disk storage shape stays user-centric (`users[]`,
+ * `StoredUser`) so a future cli-core extraction is a drop-in.
+ */
+export class AccountNotFoundError extends CliError {
+    constructor(ref: UserRef) {
+        super(
+            'ACCOUNT_NOT_FOUND',
+            `No stored account matches "${ref}". Use \`tw account list\` to see authenticated accounts.`,
```
doistbot, line 28: [P2] These new hints point users to `tw account ...` and `--user`, but neither surface exists in this PR. That leaves the CLI recommending commands/flags users cannot actually run, which conflicts with the repo guidance to keep user-facing command references accurate. Please rewrite the hints to only mention currently shipped flows, or defer these messages until the follow-up PR lands.

src/lib/auth.ts
```diff
@@ -19,306 +41,488 @@ export class NoTokenError extends CliError {
...
-            if (cleanupWarning) {
-                warn(cleanupWarning)
-            }
-        } catch (error) {
-            if (!(error instanceof SecureStoreUnavailableError)) {
-                throw error
-            }
-        }
+    const target = resolveTargetUser(config, opts.ref)
+    if (target.kind === 'legacy') {
```
doistbot, line 124: [P2] The stored-auth behavior changes here (legacy configs now fall back instead of auto-migrating, and auth is now per-user under the hood), but there’s no matching `SKILL_CONTENT` / generated skill update in the diff. Repo guidance requires the skill docs to stay in sync when auth command behavior changes, and the current setup section still describes automatic legacy-token migration.

scottlovegrove:

PR #218, opened 2026-05-11
feat(auth): [2/4] add global --user flag and tw account list/use commands

scottlovegrove:
**Part 2 of 4 stacked PRs.** Builds on #217.

When all 4 land, the integration branch `scottl/multi-user-base` merges into `main` as one release.

## Order

- #217 — refactor: foundation
- **#218 this PR** — feat: `--user` flag and `tw account list/use`
- #219 — fix: per-account `current_workspace`
- #220 — feat: postinstall v1 → v2 migration

## Summary

Exposes the user-facing surface on top of part 1's storage layer.

### Global `--user <id|email>` (`lib/global-args.ts`, `src/index.ts`)

Selects which stored account this invocation runs against. `undefined` falls back to default / single-stored.

- `parseTwLocalFlags` extends with `user`. The flag honours `--` (tokens after the terminator are positional).
- `stripUserFlag` removes `--user <ref>` from argv before Commander sees it, since Commander has no global-option attachment.
- `validateUserFlag` runs pre-Commander to reject:
  - bare `--user` (no value)
  - `--user=` (empty value)
  - `--user <known-subcommand>` (typo: forgot the value)
  Index.ts surfaces the error as `USER_FLAG_INVALID` CliError.

### `tw account list / use` (`commands/account/`)

- `account list` shows all stored accounts with a default marker. `--json` emits the full record. Named `account` because `tw user` already shows current Twist API user info.
- `account use <ref>` calls `setDefaultUserId` and confirms.

### `tw auth status` (`commands/auth/status.ts`)

- Reads config + metadata in parallel; renders an env-source marker for `TWIST_API_TOKEN` and a default marker for the active user.
- Lists other stored accounts when more than one is present.
- `--json` envelope grows `source`, `isDefault`, and `storedAccounts[]`.

### `tw auth logout` (`commands/auth/logout.ts`)

- Now respects `--user <ref>` so multi-user installs can scope the logout. Passes the ref through to `clearApiToken`.

### SKILL_CONTENT

Documents `--user` + `tw account` subcommands. `skills/twist-cli/SKILL.md` regenerated via `npm run sync:skill`.

### Tests

- New `global-args.user.test.ts` covers `parseGlobalArgs` / `stripUserFlag` / `validateUserFlag` for the bare / empty / subcommand-lookalike / terminator cases.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 556 pass
[...]

PR #219, opened 2026-05-11
fix(account): [3/4] scope current_workspace per-account

scottlovegrove:
**Part 3 of 4 stacked PRs.** Builds on #218.

## Order

- #217 — refactor: foundation
- **#219 this PR** — fix: per-account `current_workspace`
- #220 — feat: postinstall v1 → v2 migration

## Summary

With multiple authenticated accounts in play (part 2 wired up `--user`), the active workspace id must travel with each account — previously it lived at the top level of the config file and was shared across every account. `tw --user other inbox` swapped the token correctly but kept the previous account's workspace id, which the other account may not be a member of (API returned a 4xx body in place of an array, downstream `.map` blew up).

### `api.ts`

- `getTwistClient` resolves the active account via `resolveActiveUser({ ref: getRequestedUserRef() })` so every API call honours `--user`.
- `getCurrentWorkspaceId` reads from the active `StoredUser.current_workspace` and persists fresh selections back onto the same record. Env-token / legacy fallback flows resolve freshly from the session user each call (no persistent slot).
- `setActiveAccountWorkspace` writes the workspace selection onto the active account record. Used by `tw workspace use`.
- A one-shot lazy migration moves any existing top-level `currentWorkspace` onto the default (or only) account on the next workspace lookup, and drops it otherwise.

### `workspace.ts`

- `tw workspace use <ref>` now calls `setActiveAccountWorkspace` instead of writing to the top-level config.

### `inbox.ts` and `channel/threads.ts` — defensive assertBatchData

Route batch responses through the existing `assertBatchData` so a future API error surfaces as `BATCH_FAILED` CliError instead of a `TypeError` on `.map`. Defensive against API changes that break the SDK's batch-response schema validation — and exactly what caused the original symptom on the multi-account branch.

### Tests

- `inbox.test.ts` and `channel/threads.test.ts` mocks updated to include `assertBatchData` (pass-through implementation matching the production guard).

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 556 pass

### DIFF
--- a/src/commands/channel/threads.test.ts
+++ b/src/commands/channel/threads.test.ts
@@ -14,6 +14,12 @@ const refsMocks = vi.hoisted(() => ({
 vi.mock('../../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
+        if ((response.code ?? 200) >= 400 || response.data == null) {
+            throw new Error(`Failed to fetch ${label}.`)
+        }
+        return response.data
+    },
 }))
 
 vi.mock('../../lib/refs.js', () => ({
--- a/src/commands/inbox.test.ts
+++ b/src/commands/inbox.test.ts
@@ -10,6 +10,12 @@ const apiMocks = vi.hoisted(() => ({
 vi.mock('../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
+        if ((response.code ?? 200) >= 400 || response.data == null) {
+            throw new Error(`Failed to fetch ${label}.`)
+        }
+        return response.data
+    },
 }))
 
 vi.mock('../lib/refs.js', () => ({
--- a/src/lib/api.ts
+++ b/src/lib/api.ts
@@ -5,12 +5,14 @@ import {
     type Workspace,
     type WorkspaceUser,
 } from '@doist/twist-sdk'
-import { getApiToken } from './auth.js'
-import { getConfig, updateConfig } from './config.js'
+import { resolveActiveUser } from './auth.js'
+import { type Config, getConfig, setConfig, type StoredUser } from './config.js'
 import { CliError, isInsufficientScope } from './errors.js'
+import { getRequestedUserRef } from './global-args.js'
 import { ensureWriteAllowed, isMutatingMethod } from './permissions.js'
 import { getProgressTracker } from './progress.js'
 import { withSpinner } from './spinner.js'
+import { getStoredUsers, updateStoredUser } from './users.js'
 
 // Mapping of API method paths to user-friendly spinner messages
 const API_SPINNER_MESSAGES: Record<string, { text: string; color?: 'blue' | 'green' | 'yellow' }> =
@@ -237,8 +239,8 @@ export function createWrappedTwistClient(token: string): TwistApi {
 
 export async function getTwistClient(): Promise<TwistApi> {
     if (!apiClient) {
-        const token = await getApiToken()
-        apiClient = createWrappedTwistClient(token)
+        const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
+        apiClient = createWrappedTwistClient(resolved.token)
     }
     return apiClient
 }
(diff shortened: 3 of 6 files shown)


doistbot (commented):
This PR effectively scopes the active workspace to individual accounts and strengthens batch response handling to support the multi-account transition. The new architecture and defensive checks nicely improve the overall robustness of the CLI. A few refinements are needed to prevent concurrent configuration overwrites, ensure proper handling of legacy workspaces and environment tokens, optimize redundant keychain reads, and improve test coverage for the new migration and error paths.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F219&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad1b8b770-4d67-11f1-916e-d5de67e3dd65%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad1b8b770-4d67-11f1-916e-d5de67e3dd65&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
     }
+    return workspaces[0].id
+}
 
-    const defaultWorkspace = workspaces[0]
-    await updateConfig({ currentWorkspace: defaultWorkspace.id })
-    return defaultWorkspace.id
+async function persistUserWorkspace(userId: string, workspaceId: number): Promise<void> {
+    const config = await getConfig()
+    await setConfig(updateStoredUser(config, userId, { current_workspace: workspaceId }))
```
doistbot, line 319: [P1] `setConfig` replaces the entire config file locklessly. Any concurrent updates to other top-level keys (like `userSettings`) made between `getConfig()` and `setConfig()` will be overwritten and lost.

Consider restoring the `updateConfig` import from `./config.js` and using it to preserve cli-core's read-merge-write atomicity for top-level keys (the same applies to the migration step on line 286):
```typescript
    const config = await getConfig()
    const updated = updateStoredUser(config, userId, { current_workspace: workspaceId })
    await updateConfig({ users: updated.users })
```

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
+    const { currentWorkspace: _drop, ...rest } = config
+    return { ...rest, users: next }
+}
+
+/**
+ * Persist the active user's workspace selection. Used by `tw workspace use`.
+ */
+export async function setActiveAccountWorkspace(workspaceId: number): Promise<void> {
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
+    if (!resolved.id) return
```
doistbot, line 353: [P2] Returning silently here means `tw workspace use <ref>` with an env-token will print a success message but silently fail to persist the selection. Consider throwing a `CliError` so the user knows this command has no effect in env-token flows:
```typescript
    if (!resolved.id) {
        throw new CliError('NOT_SUPPORTED', 'Cannot persist workspace selection when using an environment token.')
    }
```

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
     }
 
-    const config = await getConfig()
-    if (config.currentWorkspace) {
-        return config.currentWorkspace
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
+
+    // Env-token / legacy-fallback flows have no persistent slot to write
+    // into; resolve the workspace freshly from the API each call.
+    if (!resolved.id) {
```
doistbot, line 273: [P1] This `!resolved.id` branch also catches the still-supported v1 fallback user from `resolveLegacyToken()`. On those installs it skips the persisted legacy `currentWorkspace` entirely, so users who had selected a non-default workspace get silently pushed back to `defaultWorkspace`/the first workspace. Please preserve the legacy `currentWorkspace` read here (or migrate it) until the v1→v2 migration is guaranteed to have run.

src/commands/channel/threads.test.ts
```diff
@@ -14,6 +14,12 @@ const refsMocks = vi.hoisted(() => ({
 vi.mock('../../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
```
doistbot, line 17: [P2] Instead of duplicating `assertBatchData` and introducing drift from the production version (e.g., `Error` vs `CliError`, `code?:` vs `code:`), reuse the actual implementation via `importOriginal`:
```typescript
vi.mock('../../lib/api.js', async (importOriginal) => ({
    ...(await importOriginal<typeof import('../../lib/api.js')>()),
    getTwistClient: apiMocks.getTwistClient,
    getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
}))
```

src/commands/inbox.test.ts
```diff
@@ -10,6 +10,12 @@ const apiMocks = vi.hoisted(() => ({
 vi.mock('../lib/api.js', () => ({
     getTwistClient: apiMocks.getTwistClient,
     getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+    assertBatchData: <T>(response: { code?: number; data: T }, label: string): T => {
```
doistbot, line 13: [P2] As with `threads.test.ts`, this mock re-implements `assertBatchData`. Use `importOriginal` to retain the actual implementation from `../lib/api.js` and eliminate this duplicated logic.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
     }
 
     const sessionUser = await getSessionUser()
     if (sessionUser.defaultWorkspace) {
-        await updateConfig({ currentWorkspace: sessionUser.defaultWorkspace })
+        await persistUserWorkspace(userId, sessionUser.defaultWorkspace)
         return sessionUser.defaultWorkspace
     }
 
+    const id = await firstWorkspaceId()
```
doistbot, line 302: [P2] The workspace fallback logic here (`getSessionUser()`, checking `defaultWorkspace`, and invoking `firstWorkspaceId()`) is identical to the unauthenticated early exit on line 273. These can be unified into a single resolution path at the end of the function, conditionally calling `persistUserWorkspace(resolved.id, id)` if an ID is present.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
+    await setConfig(updateStoredUser(config, userId, { current_workspace: workspaceId }))
+}
+
+function migrateLegacyCurrentWorkspace(config: Config): Config {
+    if (typeof config.currentWorkspace !== 'number') return config
+    const users = getStoredUsers(config)
+    if (users.length === 0) {
+        // No users to attach to yet — strip the orphan so it stops being
+        // surfaced by `tw doctor` and friends.
+        const { currentWorkspace: _drop, ...rest } = config
```
doistbot, line 328: [P3] The logic to strip the orphan workspace (`const { currentWorkspace: _drop, ...rest } = config`) is duplicated across three different branches. Destructuring `currentWorkspace` once at the beginning of the function allows you to share the `rest` reference and simplify these returns.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
+        return rest
+    }
+    const targetId = config.user?.default_user ?? (users.length === 1 ? users[0].id : undefined)
```
doistbot, line 331: [P2] This reimplements default-user selection with the raw `default_user` id instead of the repo's normal "resolved default or single stored user" rules. If `user.default_user` is stale/orphaned and there's exactly one real stored user, `targetId` becomes the orphan id, no user gets `current_workspace`, and the legacy value is dropped. Please verify the default id exists (or reuse the shared helper) before preferring it over the sole stored user.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
         return flagValue
     }
 
-    const config = await getConfig()
-    if (config.currentWorkspace) {
-        return config.currentWorkspace
+    const resolved = await resolveActiveUser({ ref: getRequestedUserRef() })
```
doistbot, line 269: [P2] Calling `resolveActiveUser` here forces an API token lookup via the OS credential manager. Since `getTwistClient` (and `setActiveAccountWorkspace`) also calls it, commands will trigger duplicate, slow keychain reads. Consider wrapping the `resolveActiveUser` call in a module-level cached promise so the resolved user can be shared across calls in the same process without redundant overhead.

src/lib/api.ts
```diff
@@ -264,27 +266,92 @@ export async function getCurrentWorkspaceId(flagValue?: number): Promise<number>
...
 
-    const defaultWorkspace = workspaces[0]
-    await updateConfig({ currentWorkspace: defaultWorkspace.id })
-    return defaultWorkspace.id
+async function persistUserWorkspace(userId: string, workspaceId: number): Promise<void> {
+    const config = await getConfig()
+    await setConfig(updateStoredUser(config, userId, { current_workspace: workspaceId }))
+}
+
+function migrateLegacyCurrentWorkspace(config: Config): Config {
```
doistbot, line 322: [P2] The legacy workspace migration is a high-value scenario introduced by this PR, but it is currently untested. Given the complex branching logic (dropping orphans, resolving defaults, and mapping user records), it should have dedicated unit tests. Consider exporting this pure function or moving it to `config.ts` so its edge cases can be directly tested.

src/commands/inbox.test.ts
doistbot, line 13: [P2] These suites now stub `assertBatchData`, but they still don’t exercise the new failure branch that this change is defending. Please add a regression test here (and the parallel one in `channel/threads.test.ts`) where `batch()` returns something like `{ code: 404, data: null }` and assert the command surfaces the clear `Failed to fetch …` error instead of blowing up on `.map`.

src/lib/api.ts
doistbot, line 319: [P2] This new `setConfig(...)` call (and the migration write above) bypasses the repo’s structured error-handling convention: if the config write fails, commands will now leak a raw filesystem/cli-core exception instead of a `CliError('CONFIG_WRITE_FAILED', ...)`. Please wrap these new config writes the same way `auth.ts` does so user-facing failures stay consistent in text and `--json` mode.

scottlovegrove:

PR #220, opened 2026-05-11
feat(auth): [4/4] postinstall v1 → v2 migration

scottlovegrove:
**Part 4 of 4 stacked PRs — final piece.** Builds on #219.

When this lands, the integration branch `scottl/multi-user-base` is ready to merge into `main` as one release.

## Order

- #217 — refactor: foundation
- #219 — fix: per-account `current_workspace`
- **#220 this PR** — feat: postinstall v1 → v2 migration

## Summary

With multi-user storage in place (parts 1-3), v1 single-user installs need a path forward. This adds a best-effort postinstall migration so existing tokens move to the new per-user keyring slot transparently on upgrade.

### `lib/migrate-auth.ts` (NEW)

- `migrateLegacyAuth` is best-effort: any failure (offline, invalid token, keyring unavailable) leaves the v1 state untouched so the runtime fallback in `resolveActiveUser` keeps serving the legacy token until the next attempt or `tw auth login`.
- Detects v1 shape via absence of `config.users`. `users: []` (a clean v2 logout state) is treated as already-migrated — important so a logged-out v2 install can't silently re-auth from a stale legacy keyring entry.
- Loads the legacy token (config plaintext first, then `api-token` keyring slot) and identifies the user via a one-shot REST call to `/api/v3/users/get_session_user`. Uses raw `fetch` so the postinstall path doesn't pull in the SDK.
- On success: writes a `StoredUser` to `config.users[0]`, sets `config.user.default_user`, stores the token in `user-<id>`, deletes the legacy `api-token` slot (best effort), and strips legacy top-level fields.
- On secure-store failure: persists the token to `users[0].api_token` as plaintext fallback (matches the `upsertUser` contract).

### `postinstall.ts`

Calls `migrateLegacyAuth({ silent: true })`. Failures are swallowed so a flaky upgrade never blocks `npm install`.

### Tests

- `migrate-auth.test.ts` covers: already-migrated short-circuit (`users: []` and `users: [...]`), no-legacy-state empty config, successful migration of a plaintext + keyring legacy token, and the skipped-on-401 path (config untouched).
- `postinstall.test.ts` asserts `migrateLegacyAuth({ silent: true })` is invoked and that errors are swallowed.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 563 pass

## After this lands

Open a final PR `scottl/multi-user-base` → `main` so semantic-release ships one minor (the `feat:` commits across this stack add up to one release with combined release notes).

### DIFF
--- a/src/lib/migrate-auth.test.ts
+++ b/src/lib/migrate-auth.test.ts
@@ -1,0 +1,135 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
+
+const configState: { current: import('./config.js').Config } = { current: {} }
+
+vi.mock('./config.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('./config.js')>()
+    return {
+        ...actual,
+        getConfig: vi.fn(async () => configState.current),
+        setConfig: vi.fn(async (next: import('./config.js').Config) => {
+            configState.current = next
+        }),
+    }
+})
+
+vi.mock('./secure-store.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('./secure-store.js')>()
+    return {
+        ...actual,
+        createSecureStore: vi.fn(),
+    }
+})
+
+import type { Config } from './config.js'
+import { migrateLegacyAuth } from './migrate-auth.js'
+import { createSecureStore } from './secure-store.js'
+
+const mockCreateSecureStore = vi.mocked(createSecureStore)
+
+function memorySecureStore(initial: string | null = null) {
+    let secret = initial
+    return {
+        getSecret: vi.fn(async () => secret),
+        setSecret: vi.fn(async (v: string) => {
+            secret = v
+        }),
+        deleteSecret: vi.fn(async () => {
+            const had = secret !== null
+            secret = null
+            return had
+        }),
+    }
+}
+
+function fetchOk(body: object) {
+    return vi.fn(async () => new Response(JSON.stringify(body), { status: 200 }))
+}
+
+function setConfigState(next: Config) {
+    configState.current = next
+}
+
+describe('migrateLegacyAuth', () => {
+    let consoleSpy: ReturnType<typeof vi.spyOn>
+
+    beforeEach(() => {
+        setConfigState({})
+        mockCreateSecureStore.mockImplementation(() => memorySecureStore())
+        consoleSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
+    })
+
+    afterEach(() => {
+        consoleSpy.mockRestore()
+    })
+
+    it('returns already-migrated when users key is present (even if empty)', async () => {
+        setConfigState({ users: [] })
+        expect(await migrateLegacyAuth({ silent: true })).toEqual({ status: 'already-migrated' })
+    })
+
+    it('returns no-legacy-state when config is completely empty', async () => {
+        setConfigState({})
+        expect(await migrateLegacyAuth({ silent: true })).toEqual({ status: 'no-legacy-state' })
+    })
+
+    it('migrates legacy plaintext token into a per-user record', async () => {
+        const legacyStore = memorySecureStore()
+        const userStore = memorySecureStore()
+        mockCreateSecureStore.mockImplementation((slot?: string) =>
+            slot === 'user-42' ? userStore : legacyStore,
+        )
+        setConfigState({
+            token: 'legacy-token',
+            authMode: 'read-write',
+            authScope: 'user:read threads:read',
+        })
+
+        const fetchImpl = fetchOk({ id: 42, email: 'ada@example.com', name: 'Ada' })
+        const result = await migrateLegacyAuth({ silent: true, fetchImpl })
+
+        expect(result).toMatchObject({
+            status: 'migrated',
+            migratedUserId: '42',
+            migratedEmail: 'ada@example.com',
+        })
+        expect(configState.current.users).toEqual([
+            {
+                id: '42',
+                email: 'ada@example.com',
+                name: 'Ada',
+                auth_mode: 'read-write',
+                auth_scope: 'user:read threads:read',
+            },
+        ])
+        expect(configState.current.user?.default_user).toBe('42')
+        expect(configState.current.token).toBeUndefined()
+        expect(configState.current.authMode).toBeUndefined()
+        expect(configState.current.config_version).toBe(2)
+        expect(userStore.setSecret).toHaveBeenCalledWith('legacy-token')
+    })
+
+    it('migrates legacy keyring token (no plaintext) via the api-token slot', async () => {
+        const legacyStore = memorySecureStore('keyring-token')
+        const userStore = memorySecureStore()
+        mockCreateSecureStore.mockImplementation((slot?: string) =>
+            slot === 'user-7' ? userStore : legacyStore,
+        )
+        setConfigState({})
+        // Make config look v1 by NOT having `users` key, but with no plaintext
+        // token. The keyring read happens via createSecureStore(LEGACY_*).
+        const fetchImpl = fetchOk({ id: 7, email: 'k@example.com' })
+        const result = await migrateLegacyAuth({ silent: true, fetchImpl })
+        expect(result.status).toBe('migrated')
+        expect(configState.current.users?.[0].id).toBe('7')
+    })
+
+    it('skips and leaves v1 state untouched when getSessionUser fails', async () => {
+        setConfigState({ token: 'bad-token' })
+        const failingFetch = vi.fn(async () => new Response('nope', { status: 401 }))
+        const result = await migrateLegacyAuth({ silent: true, fetchImpl: failingFetch })
+        expect(result.status).toBe('skipped')
+        expect(configState.current.token).toBe('bad-token')
+        expect(configState.current.users).toBeUndefined()
+    })
+})
--- a/src/lib/migrate-auth.ts
+++ b/src/lib/migrate-auth.ts
@@ -1,0 +1,184 @@
+/**
+ * One-time migration of v1 single-user auth state into the v2 multi-user
+ * shape. Invoked from `src/postinstall.ts` so existing installs upgrade
+ * transparently. Best-effort: any failure (offline, invalid token, keyring
+ * unavailable) leaves the v1 state untouched so the runtime fallback in
+ * `resolveActiveUser` can keep serving the legacy token until the next
+ * upgrade attempt or `tw auth login`.
+ */
+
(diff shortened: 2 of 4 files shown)


src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+async function loadLegacyToken(config: Config): Promise<string | null> {
+    if (typeof config.token === 'string' && config.token.trim()) {
+        return config.token.trim()
+    }
+    try {
+        const legacyStore = createSecureStore(LEGACY_ACCOUNT_NAME)
+        const stored = await legacyStore.getSecret()
+        if (stored?.trim()) return stored.trim()
+    } catch (error) {
+        if (!(error instanceof SecureStoreUnavailableError)) throw error
```
doistbot, line 136: [P1] Swallowing `SecureStoreUnavailableError` here causes `loadLegacyToken` to return `null` when the keyring is locked or unavailable. `migrateLegacyAuth` will then incorrectly assume there is "no usable token" and proceed to wipe the legacy config (`authMode`, etc.) and write `users: []`. This violates the PR description that a keyring failure leaves the v1 state untouched, and will permanently log out the user. Consider letting the error throw (and optionally catching it in `migrateLegacyAuth` to return `skipped(...)`) so the migration is safely aborted.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    return { status: 'migrated', migratedUserId: user.id, migratedEmail: user.email }
+}
+
+async function loadLegacyToken(config: Config): Promise<string | null> {
+    if (typeof config.token === 'string' && config.token.trim()) {
+        return config.token.trim()
+    }
+    try {
+        const legacyStore = createSecureStore(LEGACY_ACCOUNT_NAME)
```
doistbot, line 132: [P1] This keyring read should be gated on `config.pendingSecureStoreClear`. That flag is the legacy “logged out, but we couldn't clear the secure store” marker; `resolveActiveUser()` intentionally refuses to read the old `api-token` slot when it is set. As written, postinstall can migrate a stale legacy secret back into `users[0]` and silently log a previously logged-out user back in.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+    // Identify the user behind the legacy token via a one-shot REST call —
+    // no SDK import to keep postinstall lightweight.
+    let user: SessionUser
+    try {
+        user = await fetchSessionUser(legacyToken, fetchImpl)
+    } catch (error) {
+        return skipped(opts, `could not identify Twist user (${describe(error)})`)
+    }
+
+    const record: StoredUser = {
```
doistbot, line 80: [P2] Missed opportunity to reuse `upsertUser` from `src/lib/auth.ts`. This block duplicates the token-saving behavior, plaintext fallback, `default_user` assignment, and legacy field stripping of `upsertUser`. It also misses the rollback logic `upsertUser` uses to delete the newly-created secure store entry if `setConfig` fails. Since `auth.ts` does not pull in the SDK (`api.ts`), it is safe to use in postinstall. Consider replacing lines 80-110 with a `try/catch` around a call to `upsertUser`.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    if (!opts.silent) {
+        console.error(`twist-cli: migrated existing token to multi-user store (${user.email}).`)
+    }
+
+    return { status: 'migrated', migratedUserId: user.id, migratedEmail: user.email }
+}
+
+async function loadLegacyToken(config: Config): Promise<string | null> {
```
doistbot, line 127: [P2] `loadLegacyToken()` duplicates the legacy-credential lookup already implemented by `resolveLegacyToken()` in `auth.ts` (config plaintext first, then the legacy `api-token` keyring slot, while tolerating `SecureStoreUnavailableError`). That gives the migration path and the runtime fallback two separate copies of the same source-of-truth logic. Please extract a shared legacy-token reader and have both callers use it so they can't drift apart.

src/lib/migrate-auth.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    it('skips and leaves v1 state untouched when getSessionUser fails', async () => {
+        setConfigState({ token: 'bad-token' })
+        const failingFetch = vi.fn(async () => new Response('nope', { status: 401 }))
+        const result = await migrateLegacyAuth({ silent: true, fetchImpl: failingFetch })
+        expect(result.status).toBe('skipped')
+        expect(configState.current.token).toBe('bad-token')
+        expect(configState.current.users).toBeUndefined()
+    })
+})
```
doistbot, line 135: [P2] The PR description mentions "On secure-store failure: persists the token to `users[0].api_token` as plaintext fallback", but this fallback path is missing from the test suite. Consider adding a test where `userStore.setSecret` rejects with a `SecureStoreUnavailableError` to ensure the plaintext fallback behaves correctly.

src/postinstall.test.ts
```diff
@@ -15,9 +19,21 @@ describe('postinstall', () => {
         expect(updateAllInstalledSkills).toHaveBeenCalledWith({ local: false })
     })
 
-    it('swallows errors', async () => {
+    it('calls migrateLegacyAuth silently', async () => {
+        const { migrateLegacyAuth } = await import('./lib/migrate-auth.js')
+        await import('./postinstall.js')
+        expect(migrateLegacyAuth).toHaveBeenCalledWith({ silent: true })
```
doistbot, line 25: [P2] This assertion can pass on a call made by an earlier test. The first test imports `postinstall.js`, which already calls `migrateLegacyAuth`, and `beforeEach` only resets modules—not mock history. Clear/reset mocks between tests (or assert from a zeroed mock) so this case proves that this import triggers migration.

src/postinstall.test.ts
```diff
@@ -15,9 +19,21 @@ describe('postinstall', () => {
...
+    it('swallows skill-update errors', async () => {
         const { updateAllInstalledSkills } = await import('./lib/skills/update-installed.js')
         vi.mocked(updateAllInstalledSkills).mockRejectedValueOnce(new Error('fail'))
         await expect(import('./postinstall.js')).resolves.not.toThrow()
     })
+
+    it('swallows migration errors', async () => {
+        const { migrateLegacyAuth } = await import('./lib/migrate-auth.js')
+        vi.mocked(migrateLegacyAuth).mockRejectedValueOnce(new Error('fail'))
+        await expect(import('./postinstall.js')).resolves.not.toThrow()
```
doistbot, line 37: [P2] This only verifies that importing `postinstall.js` doesn't reject; it never proves that the rejected `migrateLegacyAuth` promise was actually created. If the migration call were removed entirely, this test would still pass. Add an assertion that `migrateLegacyAuth({ silent: true })` was invoked (with mock state cleared first) so the test covers swallowing that specific failure path.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+        const stored = await legacyStore.getSecret()
+        if (stored?.trim()) return stored.trim()
+    } catch (error) {
+        if (!(error instanceof SecureStoreUnavailableError)) throw error
+    }
+    return null
+}
+
+async function fetchSessionUser(token: string, fetchImpl: typeof fetch): Promise<SessionUser> {
+    const response = await fetchImpl(SESSION_ENDPOINT, {
```
doistbot, line 142: [P1] The [Service Production Readiness standard](https://handbook.doist.com/doc/standard-service-production-readiness-qgNRng8TJx) mandates applying timeouts to all network calls. Because `fetch` has no default timeout, a hanging network connection to the API will cause this `postinstall` script to block the user's `npm install` indefinitely. Pass an `AbortSignal.timeout(...)` in the fetch options to ensure the migration fails fast if the network is unresponsive.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,184 @@
...
+    // Clean up the legacy keyring entry — best effort, never fatal.
+    try {
+        const legacyStore = createSecureStore(LEGACY_ACCOUNT_NAME)
+        await legacyStore.deleteSecret()
+    } catch {
+        // ignore
+    }
+
+    if (!opts.silent) {
+        console.error(`twist-cli: migrated existing token to multi-user store (${user.email}).`)
```
doistbot, line 121: [P2] This success log prints `user.email` to stderr. Doist's observability standard says logs must not include PII, so please log the migrated user id (or just that migration succeeded) instead of the email address. https://handbook.doist.com/doc/standard-observability-vWRaOfitho

doistbot (commented):
This final PR in the multi-user stack introduces a postinstall migration script to smoothly transition existing v1 auth tokens to the new v2 per-user secure store. The best-effort approach to keeping `npm install` unblocked while upgrading credentials in the background is a great UX choice that cleanly rounds out the multi-user feature. There are a few important edge cases to refine to ensure the migration is truly safe and robust, specifically around avoiding accidental logouts during keyring failures, preventing hanging network calls, respecting legacy logged-out states, and omitting PII from logs, alongside some opportunities to consolidate the token management logic and tighten the test assertions.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F220&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Add9229f0-4d67-11f1-9af3-5ed21156c769%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Add9229f0-4d67-11f1-9af3-5ed21156c769&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

scottlovegrove:

PR #222, opened 2026-05-12
refactor(auth): integrate @doist/cli-core 0.10.0 (logout + status)

scottlovegrove:
## Summary

Bumps `@doist/cli-core` to **0.10.0** and migrates `tw auth logout` + `tw auth status` onto the new `attachLogoutCommand` / `attachStatusCommand` registrars that ship alongside the existing `attachLoginCommand`. Mirrors [Doist/todoist-cli#334](https://github.com/Doist/todoist-cli/pull/334) for `twist-cli`.

- `TwistTokenStore` gains `getLastStorageResult()` + `getLastClearResult()` methods (renamed from the prior `lastSaveResult` getter) so the logout `onCleared` callback can surface keyring-fallback warnings — cli-core's `TokenStore.clear: void` contract can't carry the `TokenStorageResult` directly.
- `auth/index.ts` now creates a single shared `TwistTokenStore` instance used by login, logout, and status (login's wrapper takes the store as a parameter instead of constructing its own).
- `attachTwistStatusCommand` routes both the active-snapshot path (`fetchLive`) and the unauthenticated path (`onNotAuthenticated`) through one `gatherStatusData` helper so env-token mode and the missing-identity path (both return `null` from `TokenStore.active()` per the adapter's documented contract) render identically to the persisted-default-user path. 401-translation lives inside the gather helper so both branches share it.
- `auth logout` and `auth status` now accept `--ndjson` for free (framework-registered) on top of `--json`. `auth status --json` expands its payload from `{id, email, name}` to `{id, email, name, authMode, authScope, source}` for parity with todoist-cli's expanded envelope. `auth logout --json` emits `{"ok": true}`; `--ndjson` stays silent.
- `auth token` stays hand-rolled — it remains a positional save/prompt flow that doesn't fit the cli-core registrar surface.

## Test plan

- [x] \`npm run lint:check\` (oxlint + oxfmt) clean
- [x] \`npm run type-check\` clean
- [x] \`npm test\` — 586/586 across 37 files
- [x] \`npm run build\` clean
- [x] \`npm run check:skill-sync\` clean
- [x] Smoke: \`HOME=\$(mktemp -d) tw auth status\` prints the legacy "No API token found." CliError envelope
- [x] Smoke: \`tw auth {logout,status} --help\` show framework-registered \`--json\` / \`--ndjson\` flags
- [ ] Manual: full OAuth login → \`tw auth status\` → \`tw auth status --json\` → \`tw auth logout\` round-trip against a real account
- [ ] Manual: \`TWIST_API_TOKEN=… tw auth status\` shows \`✓ Authenticated\` env-mode branch (source = \`env\`)
- [ ] Manual: \`tw auth logout --json\` against a real account emits \`{"ok": true}\` on stdout; keyring-fallback path routes warning to stderr

### DIFF
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -13,13 +13,20 @@ vi.mock('../../lib/auth.js', async (importOriginal) => {
 })
 
 // Mock the api module
-vi.mock('../../lib/api.js', () => ({
-    getSessionUser: vi.fn(),
-}))
+vi.mock('../../lib/api.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('../../lib/api.js')>()
+    return {
+        ...actual,
+        getSessionUser: vi.fn(),
+        createWrappedTwistClient: vi.fn(),
+    }
+})
 
 // Mock cli-core's auth subpath so login subcommand wiring doesn't drive a real
 // OAuth flow during tests. The provider + token-store units are exercised in
-// src/lib/auth-provider.test.ts.
+// src/lib/auth-provider.test.ts. attachLogoutCommand + attachStatusCommand
+// fall through to the real cli-core implementations so the integration is
+// exercised end-to-end.
 vi.mock('@doist/cli-core/auth', async (importOriginal) => {
     const actual = await importOriginal<typeof import('@doist/cli-core/auth')>()
     return {
@@ -49,9 +56,11 @@ vi.mock('chalk')
 import { createInterface, type Interface } from 'node:readline'
 import { attachLoginCommand } from '@doist/cli-core/auth'
 import { type User } from '@doist/twist-sdk'
-import { getSessionUser } from '../../lib/api.js'
+import { createWrappedTwistClient, getSessionUser } from '../../lib/api.js'
+import { type TwistAccount, type TwistTokenStore } from '../../lib/auth-provider.js'
 import { clearApiToken, getAuthMetadata, saveApiToken } from '../../lib/auth.js'
 import { registerAuthCommand } from './index.js'
+import { attachTwistStatusCommand } from './status.js'
 
 const mockCreateInterface = vi.mocked(createInterface)
 
@@ -59,6 +68,7 @@ const mockSaveApiToken = vi.mocked(saveApiToken)
 const mockClearApiToken = vi.mocked(clearApiToken)
 const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
 const mockGetSessionUser = vi.mocked(getSessionUser)
+const mockCreateWrappedTwistClient = vi.mocked(createWrappedTwistClient)
 const mockAttachLoginCommand = vi.mocked(attachLoginCommand)
 
 function createProgram() {
@@ -66,6 +76,17 @@ function createProgram() {
     program.exitOverride()
     registerAuthCommand(program)
     return program
+}
+
+const TEST_USER: User = {
+    id: 1,
+    name: 'Test User',
+    shortName: 'test',
+    bot: false,
+    timezone: 'UTC',
+    removed: false,
+    email: 'test@example.com',
+    lang: 'en',
 }
 
 describe('auth command', () => {
@@ -255,18 +276,7 @@ describe('auth command', () => {
         it('shows authenticated status when logged in', async () => {
             const program = createProgram()
 
-            const mockUser: User = {
-                id: 1,
-                name: 'Test User',
-                shortName: 'test',
-                bot: false,
-                timezone: 'UTC',
-                removed: false,
-                email: 'test@example.com',
-                lang: 'en',
-            }
-
-            mockGetSessionUser.mockResolvedValue(mockUser)
+            mockGetSessionUser.mockResolvedValue(TEST_USER)
             mockGetAuthMetadata.mockResolvedValue({
                 authMode: 'read-write',
                 source: 'config',
@@ -275,7 +285,7 @@ describe('auth command', () => {
             await program.parseAsync(['node', 'tw', 'auth', 'status'])
 
             expect(mockGetSessionUser).toHaveBeenCalled()
-            expect(consoleSpy).toHaveBeenCalledWith('✓', 'Authenticated')
+            expect(consoleSpy).toHaveBeenCalledWith('✓ Authenticated')
             expect(consoleSpy).toHaveBeenCalledWith('  Email: test@example.com')
             expect(consoleSpy).toHaveBeenCalledWith('  Name:  Test User')
             expect(consoleSpy).toHaveBeenCalledWith('  Mode:  read-write')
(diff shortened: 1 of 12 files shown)


src/commands/auth/status.ts
```diff
@@ -18,22 +49,80 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+ * resolve via `gatherStatusData(token)`.
+ */
+export function attachTwistStatusCommand(auth: Command, store: TwistTokenStore): Command {
+    let data: StatusData | null = null
 
-    console.log(chalk.green('✓'), 'Authenticated')
-    console.log(`  Email: ${user.email}`)
-    console.log(`  Name:  ${user.name}`)
-    console.log(`  Mode:  ${modeLabel}`)
+    return attachStatusCommand<TwistAccount>(auth, {
```
doistbot, line 86: [P1] Routing `tw auth status` through `attachStatusCommand` makes the command depend on `store.active()` before it can reach `onNotAuthenticated`. Our `TwistTokenStore.active()` still delegates to `probeApiToken()`, which propagates `SecureStoreUnavailableError` when the system keyring can't be opened. The old `showStatus()` path went through `getSessionUser()`/`getApiToken()`, which downgraded that case to the legacy `NO_TOKEN` message. Please catch that keyring-unavailable case in the active probe (or otherwise bypass the snapshot check) so headless/keyring-less setups don't start failing with a raw secure-store error.

src/commands/auth/status.ts
```diff
@@ -1,14 +1,45 @@
-import { TwistRequestError } from '@doist/twist-sdk'
+import { formatJson, formatNdjson } from '@doist/cli-core'
```
doistbot: [P2] By importing `formatJson` and `formatNdjson` directly from `@doist/cli-core`, this bypasses the project's standard wrappers in `../../lib/output.js`. `AGENTS.md` mandates using the local `formatJson()` wrapper for output. Update the import to pull these from `../../lib/output.js` to ensure formatting remains consistent with the rest of the CLI.

src/commands/auth/status.ts
```diff
@@ -18,22 +49,80 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+        },
+        onNotAuthenticated: async ({ view }) => {
+            // active() returned null — env-token mode, manual `tw auth token`
+            // (no persisted identity), or nothing stored. Drive the legacy
+            // resolver (getSessionUser) so all three paths render identically;
+            // getSessionUser throws NoTokenError when nothing resolves,
+            // matching prior UX.
+            data = await gatherStatusData()
+            if (view.json) {
+                console.log(formatJson(buildStatusJson(data)))
```
doistbot: [P2] This branch is reimplementing the CLI’s existing output path by pairing `console.log` with `formatJson`/`formatNdjson` directly. `src/lib/output.ts` already exposes shared `printJson` / `printNdjson` helpers (and is what other commands use), so reusing those here would avoid one-off formatting logic in `auth status` and keep future output changes centralized.

src/commands/auth/status.ts
```diff
@@ -18,22 +49,80 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+            data = await gatherStatusData(token)
+            return {
+                id: String(data.user.id),
+                label: data.user.name,
+                authMode: data.metadata.authMode,
+                authScope: data.metadata.authScope ?? '',
+            }
+        },
+        renderText: () => {
+            if (!data) throw new Error('status renderText called before fetchLive')
```
doistbot: [P2] Per `AGENTS.md`, use `CliError` from `src/lib/errors.ts` instead of bare `throw new Error(...)`. You can use `throw new CliError('INTERNAL_ERROR', '...')` here and on line 107.

src/commands/auth/status.ts
```diff
@@ -1,14 +1,45 @@
...
+ * keyring round-trip `getSessionUser()` would do via `getApiToken`.
+ *
+ * 401-translation lives here so both the snapshot path (`fetchLive`) and the
+ * legacy path (`onNotAuthenticated`) emit the same `NO_TOKEN` envelope when
+ * the token is rejected by the API.
+ */
+async function gatherStatusData(token?: string): Promise<StatusData> {
     try {
-        user = await getSessionUser()
+        const user = token
```
doistbot: [P2] `getSessionUser` (a network request) and `getAuthMetadata` (a local config read) are independent. Use `Promise.all` to fetch them concurrently instead of awaiting them sequentially.

src/commands/auth/status.ts
```diff
@@ -18,22 +49,80 @@ export async function showStatus(options: { json?: boolean }): Promise<void> {
...
+            if (!data) throw new Error('status renderJson called before fetchLive')
+            return buildStatusJson(data)
+        },
+        onNotAuthenticated: async ({ view }) => {
+            // active() returned null — env-token mode, manual `tw auth token`
+            // (no persisted identity), or nothing stored. Drive the legacy
+            // resolver (getSessionUser) so all three paths render identically;
+            // getSessionUser throws NoTokenError when nothing resolves,
+            // matching prior UX.
+            data = await gatherStatusData()
```
doistbot: [P2] This fallback now repeats token resolution for manual `tw auth token` / pre-upgrade stored tokens. `attachStatusCommand` has already called `store.active()`, and `TwistTokenStore.active()` runs `probeApiToken()` before returning `null` when no identity is persisted. Calling `gatherStatusData()` here sends `getSessionUser()` back through `getApiToken()`, so config/keyring are read twice on every such `auth status` invocation. Reuse the token discovered during the `active()` check (or return a token-only snapshot) so this path only resolves credentials once.

src/commands/auth/auth.test.ts
```diff
@@ -356,17 +445,56 @@ describe('auth command', () => {
...
+        })
+
+        it('routes warning to stderr and emits JSON envelope on stdout in --json mode', async () => {
+            const program = createProgram()
+            mockClearApiToken.mockResolvedValue(WARNING_RESULT)
+
+            await program.parseAsync(['node', 'tw', 'auth', 'logout', '--json'])
+
+            const stdoutLines = consoleSpy.mock.calls.map((c: unknown[]) => String(c[0]))
+            expect(stdoutLines).toEqual([JSON.stringify({ ok: true }, null, 2)])
```
doistbot: [P3] The exact string match against `JSON.stringify` makes this test brittle if `cli-core` ever changes its JSON indentation. Since you already upgraded the `status` JSON tests above to use `JSON.parse`, consider applying the same pattern here:
```typescript
expect(stdoutLines).toHaveLength(1)
expect(JSON.parse(stdoutLines[0])).toEqual({ ok: true })
```

doistbot (commented):
This PR successfully updates the authentication commands to leverage `@doist/cli-core` 0.10.0, neatly migrating `logout` and `status` to the new registrars while expanding JSON output capabilities. Consolidating the token store and routing status data through a shared helper noticeably improves the consistency and maintainability of the authentication flows. A few refinements are noted below to handle missing keyring scenarios gracefully, align with the project's standardized `CliError` and output formatters, optimize concurrent fetches and redundant token resolution, and strengthen test coverage for NDJSON formatting and 401 edge cases.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F222&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab0756e10-4e42-11f1-97d6-f76f325efea9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab0756e10-4e42-11f1-97d6-f76f325efea9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/auth.test.ts
```diff
@@ -284,43 +294,122 @@ describe('auth command', () => {
...
+                email: 'test@example.com',
+                name: 'Test User',
+                authMode: 'read-write',
+                authScope: 'user:read threads:read',
+                source: 'config',
+            })
         })
 
-        it('outputs JSON error when --json flag is used and not authenticated', async () => {
+        it('outputs NDJSON when --ndjson flag is used', async () => {
```
doistbot: [P2] This `--ndjson` test only `JSON.parse`s the payload, so it would still pass if the command accidentally emitted pretty JSON or any other multi-line output instead of line-delimited JSON. Please assert on the raw output too (for example, one `console.log` call whose payload contains no embedded `\n`) so the test actually locks in NDJSON behavior.

src/commands/auth/auth.test.ts
```diff
@@ -284,43 +294,122 @@ describe('auth command', () => {
...
                 'No API token found',
             )
         })
+
+        // The default tests above exercise the `onNotAuthenticated` branch (real
+        // `store.active()` resolves to null in the test env because no token /
+        // identity is persisted). This block drives a controllable snapshot
+        // store directly into `attachTwistStatusCommand` so the `fetchLive` →
+        // `renderText` / `renderJson` path is also covered.
+        describe('persisted-account snapshot path (fetchLive)', () => {
```
doistbot: [P2] The refactor moved 401 handling into `gatherStatusData()` so both the default path and this new snapshot `fetchLive` path translate an expired/invalid token into the `NO_TOKEN` `CliError`, but the new tests never exercise that branch. Right now the suite only covers happy-path snapshot output plus a generic "No API token found" error, so a regression in the shared 401 translation would slip through. Add a `TwistRequestError` 401 case here (and for the default path, or parameterize across both).

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — `TwistTokenStore.active()` now catches `SecureStoreUnavailableError` and downgrades to `null`, matching the legacy `getApiToken` behaviour so headless / keyring-less hosts fall back to the standard `NoTokenError` envelope instead of leaking the raw secure-store error.

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — dropped the direct cli-core imports entirely. The cli-core registrar handles serialization for the `fetchLive` path, and `onNotAuthenticated` simplified to `throw new NoTokenError()` so it no longer needs to format anything itself.

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — removed the hand-rolled `console.log(formatJson(...))` / `console.log(formatNdjson([...]))` branches from `onNotAuthenticated` along with the imports. The shared output path is now entirely owned by the cli-core registrar.

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — both defensive throws now use `throw new CliError(INTERNAL_ERROR, ...)`.

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — `gatherStatusData` now runs `createWrappedTwistClient(token).users.getSessionUser()` and `getAuthMetadata()` concurrently via `Promise.all`.

src/commands/auth/status.ts
scottlovegrove: Fixed in ad4898a — went with the "token-only snapshot" option. `TwistTokenStore.active()` now returns a snapshot whenever a token resolves (placeholder `id` / `label` when persisted identity is missing), so `fetchLive` reuses the probed token and credentials are only resolved once per invocation. `onNotAuthenticated` simplified to just `throw new NoTokenError()` since it now only fires when no token exists at all. Added a regression test for the env-token / no-identity path on the adapter (`active() returns a token-only snapshot with placeholder fields...`).

src/commands/auth/auth.test.ts
scottlovegrove: Fixed in ad4898a — switched to `expect(stdoutLines).toHaveLength(1)` + `expect(JSON.parse(stdoutLines[0])).toEqual({ ok: true })` so future cli-core indentation changes don't break the suite.

src/commands/auth/auth.test.ts
scottlovegrove: Fixed in ad4898a — the ndjson test now asserts `consoleSpy` was called exactly once and that the printed payload `.not.toContain(\n)`, so a pretty-printed-JSON regression would fail the test immediately.

src/commands/auth/auth.test.ts
scottlovegrove: Fixed in ad4898a — added `translates a 401 from the live API into a NO_TOKEN CliError` to the snapshot-path block. It uses the same `programWithSnapshot()` plumbing, replaces the mocked `getSessionUser` with one that rejects with `new TwistRequestError(Authentication required, 401)`, and asserts `rejects.toHaveProperty(code, NO_TOKEN)` so the shared `gatherStatusData` 401 branch is locked in.

PR #224, opened 2026-05-13
fix: guard unreadResp.data against null in channel threads command

lmjabreu:
## What

`tw channel threads <name> --json` crashed with:

```
{"error":{"code":"INTERNAL_ERROR","message":"Cannot read properties of undefined (reading 'map')"}}
```

## Root cause

`src/commands/channel/threads.ts:89` called `.map()` directly on `unreadResp.data` with no null guard:

```ts
const unreadThreadIds = new Set(unreadResp.data.map((u) => u.threadId))
```

The SDK's `client.batch()` returns `{ code, data: null }` when a sub-request fails or when `getUnread` returns no data (e.g. the user has zero unread threads in the workspace). Calling `.map()` on `null` throws.

## Fix

One-line change — add `?? []` so a null/undefined `data` falls back to an empty set, treating all threads as read:

```ts
const unreadThreadIds = new Set((unreadResp.data ?? []).map((u) => u.threadId))
```

`assertBatchData()` was intentionally **not** used here — having no unread threads is a valid state, not an error.

## Test

Added a regression test to the existing `src/commands/channel/threads.test.ts` that mocks the batch response with `data: null` for the unread sub-request and asserts the command resolves without throwing, with all threads marked `isUnread: false`.

## Verification

- `npm run type-check` ✅
- `npm test` ✅ (587 tests, 0 failures)
- Repro command: `tw channel threads "Doist Component Library" --no-spinner --json`

### DIFF
--- a/src/commands/channel/threads.test.ts
+++ b/src/commands/channel/threads.test.ts
@@ -548,6 +548,27 @@ describe('channel threads', () => {
         consoleSpy.mockRestore()
     })
 
+    it('does not crash when unreadResp.data is null (regression: batch getUnread returning null)', async () => {
+        const mockGetThreads = vi.fn().mockReturnValue('threads-descriptor')
+        const mockGetUnread = vi.fn().mockReturnValue('unread-descriptor')
+        apiMocks.getTwistClient.mockResolvedValue({
+            threads: { getThreads: mockGetThreads, getUnread: mockGetUnread },
+            batch: vi.fn().mockResolvedValue([{ data: [createThread(1)] }, { data: null }]),
+        })
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+        const program = createProgram()
+
+        await expect(
+            program.parseAsync(['node', 'tw', 'channel', 'threads', '12345', '--json']),
+        ).resolves.not.toThrow()
+
+        const output = JSON.parse(consoleSpy.mock.calls[0][0])
+        expect(output.results).toHaveLength(1)
+        expect(output.results[0].isUnread).toBe(false)
+
+        consoleSpy.mockRestore()
+    })
+
     it('--full bypasses the essential-field filter in JSON output', async () => {
         setupClient({
             threads: [createThread(1, { pinned: true })],
--- a/src/commands/channel/threads.test.ts
+++ b/src/commands/channel/threads.test.ts
@@ -11,9 +11,9 @@ const refsMocks = vi.hoisted(() => ({
     resolveChannelRef: vi.fn(),
 }))
 
-vi.mock('../../lib/api.js', () => ({
-    getTwistClient: apiMocks.getTwistClient,
-    getCurrentWorkspaceId: apiMocks.getCurrentWorkspaceId,
+vi.mock('../../lib/api.js', async (importOriginal) => ({
+    ...(await importOriginal<typeof import('../../lib/api.js')>()),
+    ...apiMocks,
 }))
 
 vi.mock('../../lib/refs.js', () => ({
@@ -89,11 +89,14 @@ function setupClient({
     unread = [],
 }: {
     threads?: Thread[]
-    unread?: { threadId: number }[]
+    unread?: { threadId: number }[] | null
 } = {}) {
     const mockGetThreads = vi.fn().mockReturnValue('threads-descriptor')
     const mockGetUnread = vi.fn().mockReturnValue('unread-descriptor')
-    const mockBatch = vi.fn().mockResolvedValue([{ data: threads }, { data: unread }])
+    const mockBatch = vi.fn().mockResolvedValue([
+        { code: 200, data: threads },
+        { code: 200, data: unread },
+    ])
 
     apiMocks.getTwistClient.mockResolvedValue({
         threads: { getThreads: mockGetThreads, getUnread: mockGetUnread },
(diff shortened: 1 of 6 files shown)


src/commands/channel/threads.ts
```diff
@@ -86,7 +86,7 @@ export async function showChannelThreads(
         client.threads.getUnread(workspaceId, { batch: true }),
     )
 
-    const unreadThreadIds = new Set(unreadResp.data.map((u) => u.threadId))
+    const unreadThreadIds = new Set((unreadResp.data ?? []).map((u) => u.threadId))
```
doistbot: [P1] While `unreadResp` is correctly guarded against `null` here, `threadsResp.data` on the next line remains unguarded and will cause the exact same `TypeError` if the `getThreads` sub-request fails (e.g., due to lost permissions or a deleted channel).

Since `threadsResp` represents the primary data rather than an optional state, consider importing `assertBatchData` from `../../lib/api.js` and using it to fail gracefully instead of crashing:

```ts
const unreadThreadIds = new Set((unreadResp.data ?? []).map((u) => u.threadId))
const threadsData = assertBatchData(threadsResp, 'threads')
let threads: DecoratedThread[] = threadsData.map((t) => ({
```

doistbot: [P1] This exact batch extraction pattern for `getUnread` is duplicated in `src/commands/inbox.ts` (`unreadData.data.map`). Please apply this same `?? []` null guard there as well to prevent the `inbox` command from crashing under identical conditions.

src/commands/channel/threads.test.ts
```diff
@@ -548,6 +548,27 @@ describe('channel threads', () => {
         consoleSpy.mockRestore()
     })
 
+    it('does not crash when unreadResp.data is null (regression: batch getUnread returning null)', async () => {
+        const mockGetThreads = vi.fn().mockReturnValue('threads-descriptor')
```
doistbot: [P2] This manually duplicates the mock setup for the SDK's batch descriptors. Consider updating the `setupClient` helper's signature to accept `null` (e.g., `unread?: { threadId: number }[] | null`) so you can reuse it here via `setupClient({ threads: [createThread(1)], unread: null })` and avoid repeating the implementation details.

doistbot (commented):
This PR nicely fixes a crash in the channel threads command by adding a necessary null guard to the unread threads response. The implementation correctly handles valid empty states, and the included regression test ensures this stability is maintained. A few additional adjustments are recommended, specifically applying this same null guard to the inbox command, using `assertBatchData` on the primary threads response to prevent a similar crash, and simplifying the test mock setup by updating the `setupClient` helper.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F224&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A22858470-4ec7-11f1-81cb-e441844c8053%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A22858470-4ec7-11f1-81cb-e441844c8053&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/channel/threads.ts
scottlovegrove: Addressed in ebc641c. `threadsResp` now goes through `assertBatchData(threadsResp, 'threads')` so a failed sub-request surfaces a clean API error instead of crashing. `unreadResp` stays tolerant via the new `getOptionalBatchData` helper (returns null on success-with-null data, throws only on real API errors) since zero unread threads is a valid empty state. Added a regression test covering the threads-error path.

scottlovegrove: Addressed in ebc641c. `src/commands/inbox.ts` was actually crashing under the same conditions after #159 was merged (it used strict `assertBatchData` on the unread response, which throws on `data: null`). Switched the inbox unread response to the new `getOptionalBatchData` helper so a workspace with zero unread threads no longer aborts `tw inbox`. Added a regression test in `src/commands/inbox.test.ts`.

src/commands/channel/threads.test.ts
scottlovegrove: Addressed in ebc641c. Updated `setupClient`'s signature to accept `unread?: { threadId: number }[] | null` and refactored the regression test to call `setupClient({ threads: [createThread(1)], unread: null })` instead of re-implementing the mock inline.

scottlovegrove:
Hey @lmjabreu — pushed a follow-up commit on top of your fix now that #159 is merged.

Quick summary of the adjustments:

- Added a `getOptionalBatchData` helper in `src/lib/api.ts` so the "null = valid empty state" pattern is reusable (and consistent with `#159`'s helpers).
- Used `assertBatchData` on `threadsResp` so a failed `getThreads` sub-request surfaces a clean error instead of crashing the same way `unreadResp` did.
- Applied the same `getOptionalBatchData` treatment to the unread response in `inbox.ts`, which was actually crashing under the same conditions after #159 landed.
- Refactored the regression test to use `setupClient({ unread: null })`, plus added tests for the new error/empty paths.
- Documented the four batch helpers (and when to use each) in `AGENTS.md`.

Functionally your original fix is preserved — the new helper just makes the pattern explicit and reusable.

scottlovegrove (approved):

frankieyan:
✅

PR #225, opened 2026-05-15
feat(auth): adopt cli-core 0.12.0 multi-user TokenStore shape

scottlovegrove:
## Summary

- Bump `@doist/cli-core` from 0.10.0 → 0.12.0.
- Reshape `TwistTokenStore` to the new uniformly-multi-user `TokenStore` contract: `active(ref?)` / `clear(ref?)` honour the optional `AccountRef`, plus new `list()` (one-element / empty) and `setDefault(ref)` (validates against the single stored account, throws `ACCOUNT_NOT_FOUND` on mismatch).
- Widen `TwGlobalArgs` to surface cli-core's new `--user` field.
- Update auth.test.ts store fakes to satisfy the wider contract.

Storage backend stays single-slot — this PR is just contract compliance so the upstream `--user <ref>` flag attached by `attachLogoutCommand` / `attachStatusCommand` / `attachTokenViewCommand` lands cleanly. Multi-account storage is a future PR.

## Test plan

- [x] `npm run type-check`
- [x] `npm run build`
- [x] `npm test` — 594 / 594 pass
- [x] `npm run lint:check`
- [ ] Manual: `tw auth status` / `tw auth logout` still work against an existing token
- [ ] Manual: `tw auth logout --user <id>` matches the stored account; mismatched ref is a no-op

### DIFF
--- a/src/lib/auth-provider.ts
+++ b/src/lib/auth-provider.ts
@@ -1,4 +1,5 @@
 import {
+    type AccountRef,
     type AuthAccount,
     type AuthProvider,
     deriveChallenge,
@@ -274,8 +275,29 @@ export type TwistTokenStore = TokenStore<TwistAccount> & {
 export function createTwistTokenStore(): TwistTokenStore {
     let lastStorageResult: TokenStorageResult | undefined
     let lastClearResult: TokenStorageResult | undefined
+    async function loadSnapshot(): Promise<{ token: string; account: TwistAccount } | null> {
+        try {
+            const { token, metadata } = await probeApiToken()
+            return {
+                token,
+                account: {
+                    id: metadata.authUserId !== undefined ? String(metadata.authUserId) : '',
+                    label: metadata.authUserName ?? '',
+                    authMode: metadata.authMode,
+                    authScope: metadata.authScope ?? '',
+                },
+            }
+        } catch (error) {
+            if (error instanceof NoTokenError) return null
+            if (error instanceof SecureStoreUnavailableError) return null
+            throw error
+        }
+    }
+    function matchesRef(account: TwistAccount, ref: AccountRef): boolean {
+        return account.id === ref || account.label === ref
+    }
     return {
-        async active() {
+        async active(ref?: AccountRef) {
             // Return a snapshot whenever a token resolves — even if the
             // persisted identity is missing (env var, manual `tw auth token`,
             // pre-upgrade config). Callers downstream (e.g. status's
@@ -290,22 +312,10 @@ export function createTwistTokenStore(): TwistTokenStore {
             // the standard `NoTokenError` envelope instead of leaking the raw
             // secure-store error — matching the legacy `getApiToken`
             // behaviour the prior `showStatus()` relied on.
-            try {
-                const { token, metadata } = await probeApiToken()
-                return {
-                    token,
-                    account: {
-                        id: metadata.authUserId !== undefined ? String(metadata.authUserId) : '',
-                        label: metadata.authUserName ?? '',
-                        authMode: metadata.authMode,
-                        authScope: metadata.authScope ?? '',
-                    },
-                }
-            } catch (error) {
-                if (error instanceof NoTokenError) return null
-                if (error instanceof SecureStoreUnavailableError) return null
-                throw error
-            }
+            const snapshot = await loadSnapshot()
+            if (!snapshot) return null
+            if (ref !== undefined && !matchesRef(snapshot.account, ref)) return null
+            return snapshot
         },
         async set(account, token) {
             const userId = Number(account.id)
@@ -316,8 +326,26 @@ export function createTwistTokenStore(): TwistTokenStore {
                 authUserName: account.label,
             })
         },
-        async clear() {
+        async clear(ref?: AccountRef) {
+            if (ref !== undefined) {
+                const snapshot = await loadSnapshot()
+                if (!snapshot || !matchesRef(snapshot.account, ref)) {
+                    lastClearResult = undefined
+                    return
+                }
+            }
             lastClearResult = await clearApiToken()
+        },
+        async list() {
+            const snapshot = await loadSnapshot()
+            return snapshot ? [{ account: snapshot.account, isDefault: true }] : []
+        },
+        async setDefault(ref: AccountRef) {
+            const snapshot = await loadSnapshot()
+            if (!snapshot || !matchesRef(snapshot.account, ref)) {
+                throw new CliError('ACCOUNT_NOT_FOUND', `No stored account matches "${ref}".`)
+            }
+            // Single-user store — already the default.
         },
         getLastStorageResult() {
             return lastStorageResult
--- a/src/lib/auth-provider.ts
+++ b/src/lib/auth-provider.ts
@@ -16,6 +16,7 @@ import {
 } from './auth.js'
 import type { AuthMode } from './config.js'
 import { CliError } from './errors.js'
+import { parseRef } from './refs.js'
 import { SecureStoreUnavailableError } from './secure-store.js'
 
 export const AUTHORIZATION_URL = 'https://twist.com/oauth/authorize'
@@ -275,7 +276,19 @@ export type TwistTokenStore = TokenStore<TwistAccount> & {
 export function createTwistTokenStore(): TwistTokenStore {
     let lastStorageResult: TokenStorageResult | undefined
     let lastClearResult: TokenStorageResult | undefined
-    async function loadSnapshot(): Promise<{ token: string; account: TwistAccount } | null> {
+
+    /**
+     * Read the stored credential. By default `NoTokenError` collapses to
+     * `null` (i.e. "nothing stored" is not an exception), while
+     * `SecureStoreUnavailableError` propagates so callers see a keyring
+     * outage as distinct from "no account". The legacy `active()` no-ref
+     * path opts in to also tolerate the outage so headless / keyring-less
+     * hosts keep falling through to the standard `NoTokenError` envelope
+     * (matching the historical `getApiToken` → `showStatus()` behaviour).
+     */
+    async function loadStoredSnapshot(
+        options: { tolerateOutage?: boolean } = {},
+    ): Promise<{ token: string; account: TwistAccount } | null> {
         try {
             const { token, metadata } = await probeApiToken()
             return {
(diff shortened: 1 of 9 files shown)


doistbot (commented):
This PR successfully bumps `@doist/cli-core` and adapts the local token store to the new multi-user `TokenStore` contract to support the upstream `--user` flag. The structural updates lay a great foundation for future multi-account support, though the updated matching logic introduces a few subtle shifts in current error reporting and logout behavior. Refining the handling of reference mismatches and storage outages, normalizing the new reference parsing to match repository standards, and rounding out the tests alongside updates to `SKILL_CONTENT` and the central error registry will ensure the new contracts are fully robust.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F225&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Acd713ff0-506b-11f1-9cb2-252fabed732f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Acd713ff0-506b-11f1-9cb2-252fabed732f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth-provider.ts
```diff
@@ -290,22 +312,10 @@ export function createTwistTokenStore(): TwistTokenStore {
...
-                    },
-                }
-            } catch (error) {
-                if (error instanceof NoTokenError) return null
-                if (error instanceof SecureStoreUnavailableError) return null
-                throw error
-            }
+            const snapshot = await loadSnapshot()
+            if (!snapshot) return null
+            if (ref !== undefined && !matchesRef(snapshot.account, ref)) return null
```
doistbot: [P1] Returning `null` on a ref mismatch makes `tw auth status --user <ref>` fall into the existing `onNotAuthenticated` path, which this repo currently maps to `NoTokenError`. With a stored token for a different account, the command will now say "No API token found" instead of reporting an account mismatch. `active()` needs a distinct mismatch signal here (for example `ACCOUNT_NOT_FOUND`) so status doesn't emit the wrong error.

src/lib/auth-provider.ts
```diff
@@ -316,9 +326,27 @@ export function createTwistTokenStore(): TwistTokenStore {
                 authUserName: account.label,
             })
         },
-        async clear() {
+        async clear(ref?: AccountRef) {
+            if (ref !== undefined) {
+                const snapshot = await loadSnapshot()
+                if (!snapshot || !matchesRef(snapshot.account, ref)) {
```
doistbot: [P1] This silent return makes a ref mismatch indistinguishable from a successful logout. `attachLogoutCommand` treats any non-throwing `clear()` as success, so `tw auth logout --user <other>` will still print `✓ Logged out` even though the token remains stored. Please surface a mismatch error here, or otherwise give the command a way to suppress the success path when nothing was cleared.

src/lib/auth-provider.ts
```diff
@@ -316,9 +326,27 @@ export function createTwistTokenStore(): TwistTokenStore {
                 authUserName: account.label,
             })
         },
-        async clear() {
+        async clear(ref?: AccountRef) {
+            if (ref !== undefined) {
+                const snapshot = await loadSnapshot()
```
doistbot: [P2] `clear(ref)` and `setDefault(ref)` now duplicate the same `loadSnapshot()` + `matchesRef()` gate that `active(ref)` already encapsulates. Reusing a shared helper (or routing these methods through `active(ref)`) would keep the single-account matching semantics in one place and avoid drift if `AccountRef` matching changes again.

src/lib/auth-provider.ts
```diff
@@ -274,8 +275,29 @@ export type TwistTokenStore = TokenStore<TwistAccount> & {
...
+                account: {
+                    id: metadata.authUserId !== undefined ? String(metadata.authUserId) : '',
+                    label: metadata.authUserName ?? '',
+                    authMode: metadata.authMode,
+                    authScope: metadata.authScope ?? '',
+                },
+            }
+        } catch (error) {
+            if (error instanceof NoTokenError) return null
+            if (error instanceof SecureStoreUnavailableError) return null
```
doistbot: [P2] `loadSnapshot()` now downgrades `SecureStoreUnavailableError` to `null`, and that helper is reused by `clear(ref)`, `list()`, and `setDefault()`. That collapses “keyring unavailable” into “no stored account”, so `auth logout --user …` becomes a silent no-op and `setDefault()` reports `ACCOUNT_NOT_FOUND` during storage outages. Please keep the tolerant fallback only for `active()`’s legacy status path, and let the ref-sensitive/listing methods preserve the storage failure distinctly.

src/lib/auth-provider.ts
```diff
@@ -274,8 +275,29 @@ export type TwistTokenStore = TokenStore<TwistAccount> & {
...
+            throw error
+        }
+    }
+    function matchesRef(account: TwistAccount, ref: AccountRef): boolean {
+        return account.id === ref || account.label === ref
```
doistbot: [P2] This bespoke string equality matcher sidesteps the repo’s normal reference parsing rules. A ref like `id:42` (and even simple case-insensitive name matching) will never match here, even though the CLI otherwise accepts normalized ref forms instead of exact raw strings. Please normalize `AccountRef` through the shared ref-parsing helpers, or add a dedicated auth-account ref parser, rather than baking ad-hoc `id`/`label` comparisons into the token store.

src/lib/auth-provider.ts
```diff
@@ -316,9 +326,27 @@ export function createTwistTokenStore(): TwistTokenStore {
...
+                    return
+                }
+            }
             lastClearResult = await clearApiToken()
         },
+        async list() {
+            const snapshot = await loadSnapshot()
+            return snapshot ? [{ account: snapshot.account, isDefault: true }] : []
+        },
+        async setDefault(ref: AccountRef) {
```
doistbot, line 384: [P2] The `setDefault` method and the new `AccountRef` matching contract (`list()`, `active(ref)`, and `clear(ref)`) are completely untested. The highest-value scenario introduced by this PR is the store's ability to validate a `ref` against the stored account. Extend `src/lib/auth-provider.test.ts` to verify that matching refs resolve correctly and mismatched refs return `null`, skip clearing, or throw `ACCOUNT_NOT_FOUND` as expected.

src/lib/global-args.ts
```diff
@@ -37,7 +37,10 @@ type TwSpecificFlags = {
  * Commander — exposing them in the type would lie about what the binary
  * supports.
  */
-export type TwGlobalArgs = Pick<CoreGlobalArgs, 'json' | 'ndjson' | 'accessible' | 'noSpinner'> &
+export type TwGlobalArgs = Pick<
+    CoreGlobalArgs,
+    'json' | 'ndjson' | 'accessible' | 'noSpinner' | 'user'
```
doistbot, line 42: [P2] The `--user` flag is now surfaced on auth commands via `cli-core`. Per the Skill Content rules in `AGENTS.md`, any new flags or options must be documented in `SKILL_CONTENT` in `src/lib/skills/content.ts`.

src/lib/auth-provider.ts
```diff
@@ -316,9 +326,27 @@ export function createTwistTokenStore(): TwistTokenStore {
...
+            const snapshot = await loadSnapshot()
+            if (!snapshot || !matchesRef(snapshot.account, ref)) {
+                throw new CliError('ACCOUNT_NOT_FOUND', `No stored account matches "${ref}".`)
```
doistbot: [P2] `ACCOUNT_NOT_FOUND` is a new error code usage, but it is not added to the centralized `ErrorCode` union in `src/lib/errors.ts`. Repository guidance calls for registering new error codes there so the CLI’s typed/documented error set stays in sync; please add it under the appropriate category.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — `active(ref)` now routes through `resolveByRef`, which throws `ACCOUNT_NOT_FOUND` on mismatch (and when nothing is stored). `tw auth status --user <other>` will surface the distinct error instead of falling into `onNotAuthenticated` / `NO_TOKEN`.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — `clear(ref)` now calls `resolveByRef` before touching storage, so a ref mismatch throws `ACCOUNT_NOT_FOUND` instead of returning silently. `attachLogoutCommand` only emits ✓ Logged out when `clear()` resolves, so the mismatched-ref case now errors out cleanly.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — `active(ref)` / `clear(ref)` / `setDefault(ref)` all route through a single `resolveByRef(ref)` helper. The match logic lives in `matchesRef` and is the only place ref semantics are encoded.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — `loadStoredSnapshot` now takes a `tolerateOutage` flag and only the legacy `active()` no-ref path opts in. `active(ref)`, `clear(ref)`, `list()`, and `setDefault(ref)` propagate `SecureStoreUnavailableError` so a keyring outage stays distinct from "no stored account". Coverage added in `auth-provider.test.ts` (active(ref) + list() outage cases).

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — `matchesRef` now runs `ref` through `parseRef` from `src/lib/refs.ts`. `id:42`, the raw `42` form, and case-insensitive label match all resolve consistently with the rest of the CLI; URL refs fall through to "no match" since they do not apply to accounts.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — added a `ref-aware lookups` describe block in `src/lib/auth-provider.test.ts` covering `active(ref)` (id / id:<n> / case-insensitive label match, mismatch → `ACCOUNT_NOT_FOUND`, NoToken → `ACCOUNT_NOT_FOUND`, keyring outage propagated), `clear(ref)` (match clears + mismatch errors without touching storage), `list()` (single-account / empty / outage propagated), and `setDefault(ref)` (match resolves silently, mismatch → `ACCOUNT_NOT_FOUND`). 13 new tests, 607/607 pass.

src/lib/global-args.ts
scottlovegrove: Fixed in 98f357f — added `--user <ref>` lines for `tw auth status`, `tw auth logout`, and `tw auth token` to the auth section of `src/lib/skills/content.ts`, then ran `npm run build && npm run sync:skill` to regenerate `skills/twist-cli/SKILL.md`.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 98f357f — added `ACCOUNT_NOT_FOUND` to the `ErrorCode` union in `src/lib/errors.ts` under the "Not found" category, sitting alongside `CHANNEL_NOT_FOUND` / `USER_NOT_FOUND` / `WORKSPACE_NOT_FOUND`.

henningmu (commented):
✅

PR #226, opened 2026-05-16
refactor(auth): adopt cli-core 0.16 createSecureStore

scottlovegrove:
## Summary

- Delete `src/lib/secure-store.ts` (+ its test) and switch every consumer to `createSecureStore` from `@doist/cli-core/auth` — cli-core's version is a strict superset (memoised `AsyncEntry` promise, `cause` propagation on `SecureStoreUnavailableError`).
- Bump `@doist/cli-core` 0.12.0 → 0.16.1.
- Keep `SECURE_STORE_DESCRIPTION` and the `serviceName` / `account` literals local in `src/lib/auth.ts` (cli-core doesn't publicly re-export the description string).
- No user-facing change. Single-user model preserved. Foundation for a later, larger migration to `createKeyringTokenStore` + `migrateLegacyAuth`.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 36 files / 605 tests pass
- [x] `npm run build`
- [ ] Manual smoke (keyring present): `tw auth login` → `tw auth status` → `tw auth logout` → `tw auth status`
- [ ] Manual smoke (keyring absent): `tw auth token <token>` falls through to config-file with the "system credential manager unavailable" warning

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #227, opened 2026-05-16
feat(auth): add `tw auth token view` via cli-core attachTokenViewCommand

scottlovegrove:
## Summary

- Wire `@doist/cli-core/auth`'s `attachTokenViewCommand` as a subcommand of the existing `tw auth token [token]` setter — `tw auth token view` prints the stored token to stdout (no envelope) so it is pipe-safe for shell composition (`eval $(tw auth token view)`, etc.).
- Mirrors the pattern todoist-cli adopted in [Doist/todoist-cli#341](https://github.com/Doist/todoist-cli/pull/341); the next standalone step in the cli-core migration after #226.
- Commander matches subcommand names before the parent action, so `tw auth token view` always dispatches to the view path — Twist OAuth tokens are opaque random strings, so a literal "view" cannot collide with a real token value.
- Refuses to print when `TWIST_API_TOKEN` is set (cli-core's `TOKEN_FROM_ENV` guard): the env var takes precedence elsewhere, so printing the saved token would disclose a secret the CLI did not manage.
- `--user <ref>` flag is added by cli-core unconditionally; matches the lone stored account via the existing `TwistTokenStore.active(ref)` path (id-or-label match through `parseRef`), so single-user semantics are preserved.

No changes to the `lib/` layer.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 5 new tests (happy path, `TOKEN_FROM_ENV` refusal, `NOT_AUTHENTICATED`, `--user` id match, `--user` id miss); 611 total pass
- [x] `npm run build`
- [x] `tw auth token --help` shows the new `view` subcommand
- [x] `tw auth token view --help` shows the `--user <ref>` flag
- [ ] Manual end-to-end: `tw auth login` → `tw auth token view` prints the bearer token → `eval $(tw auth token view)` pipes safely
- [ ] Manual env-guard: `TWIST_API_TOKEN=anything tw auth token view` exits with `TOKEN_FROM_ENV`

### DIFF
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -290,17 +290,24 @@ describe('auth command', () => {
 
         let writeSpy: ReturnType<typeof vi.spyOn>
 
+        // Capture the full stdout payload so we can assert pipe-safety: any
+        // extra preamble/trailer added by a future change would change this
+        // string and fail the test.
+        function stdoutPayload(): string {
+            return writeSpy.mock.calls.map((call: unknown[]) => String(call[0])).join('')
+        }
+
         beforeEach(() => {
             writeSpy = vi.spyOn(process.stdout, 'write').mockImplementation(() => true)
-            delete process.env[TOKEN_ENV_VAR]
         })
 
         afterEach(() => {
             writeSpy.mockRestore()
-            delete process.env[TOKEN_ENV_VAR]
+            vi.unstubAllEnvs()
         })
 
-        it('prints the stored token to stdout for pipe-safe consumption', async () => {
+        it('prints exactly the stored token to stdout with no envelope (pipe-safe)', async () => {
+            vi.stubEnv(TOKEN_ENV_VAR, '')
             mockProbeApiToken.mockResolvedValueOnce({
                 token: 'tk_stored_1234567890',
                 metadata: STORED_METADATA,
@@ -309,11 +316,12 @@ describe('auth command', () => {
             const program = createProgram()
             await program.parseAsync(['node', 'tw', 'auth', 'token', 'view'])
 
-            expect(writeSpy).toHaveBeenCalledWith('tk_stored_1234567890')
+            expect(stdoutPayload()).toBe('tk_stored_1234567890')
+            expect(consoleSpy).not.toHaveBeenCalled()
         })
 
         it('refuses to print when the env var is set so the CLI does not disclose an unmanaged token', async () => {
-            process.env[TOKEN_ENV_VAR] = 'env_token_supplied_externally'
+            vi.stubEnv(TOKEN_ENV_VAR, 'env_token_supplied_externally')
 
             const program = createProgram()
             await expect(
@@ -321,10 +329,11 @@ describe('auth command', () => {
             ).rejects.toHaveProperty('code', 'TOKEN_FROM_ENV')
 
             expect(mockProbeApiToken).not.toHaveBeenCalled()
-            expect(writeSpy).not.toHaveBeenCalled()
+            expect(stdoutPayload()).toBe('')
         })
 
         it('throws NOT_AUTHENTICATED when no token is stored', async () => {
+            vi.stubEnv(TOKEN_ENV_VAR, '')
             mockProbeApiToken.mockRejectedValueOnce(new NoTokenError())
 
             const program = createProgram()
@@ -332,10 +341,11 @@ describe('auth command', () => {
                 program.parseAsync(['node', 'tw', 'auth', 'token', 'view']),
             ).rejects.toHaveProperty('code', 'NOT_AUTHENTICATED')
 
-            expect(writeSpy).not.toHaveBeenCalled()
+            expect(stdoutPayload()).toBe('')
         })
 
         it('matches --user against the stored account by numeric id', async () => {
+            vi.stubEnv(TOKEN_ENV_VAR, '')
             mockProbeApiToken.mockResolvedValueOnce({
                 token: 'tk_stored_1234567890',
                 metadata: STORED_METADATA,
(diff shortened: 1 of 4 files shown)


src/commands/auth/index.ts
```diff
@@ -18,7 +20,21 @@ export function registerAuthCommand(program: Command): void {
...
+    // subcommand prints. Commander matches subcommand names before the parent
+    // action, so `tw auth token view` always dispatches to the view path —
+    // Twist OAuth tokens are opaque random strings so the literal "view" can
+    // never collide with a real token value.
+    const tokenCmd = auth
+        .command('token [token]')
+        .description('Save API token for CLI authentication (or use a subcommand: `view`)')
         .action(loginWithToken)
+
+    attachTokenViewCommand(tokenCmd, {
```
doistbot, line 33: [P2] Adding this subcommand also requires updating the repo’s skill reference. `src/lib/skills/content.ts` and the generated `skills/twist-cli/SKILL.md` still describe `tw auth token --user <ref>` instead of `tw auth token view [--user <ref>]`, so the agent-facing docs now instruct an invalid command. Please update the skill content and resync the generated SKILL file.

doistbot (commented):
This PR successfully introduces the `tw auth token view` subcommand to enable pipe-safe token printing via `@doist/cli-core`. It is a solid addition that mirrors established patterns while properly respecting environment variable precedence. A few adjustments are needed, specifically syncing the new command in the AI skill reference, using Vitest's `vi.stubEnv()` to prevent environment state leakage in tests, and tightening the stdout assertions to fully guarantee pipe-safe output.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F227&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A5c865410-5165-11f1-839b-9810d1ad42d4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A5c865410-5165-11f1-839b-9810d1ad42d4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/auth.test.ts
```diff
@@ -270,6 +279,89 @@ describe('auth command', () => {
...
+            authUserId: 1,
+            authUserName: 'Test User',
+            source: 'secure-store' as const,
+        }
+
+        let writeSpy: ReturnType<typeof vi.spyOn>
+
+        beforeEach(() => {
+            writeSpy = vi.spyOn(process.stdout, 'write').mockImplementation(() => true)
+            delete process.env[TOKEN_ENV_VAR]
```
doistbot: [P2] Directly deleting `process.env` variables causes state leakage. If a developer runs tests with this environment variable exported locally, this `beforeEach`/`afterEach` permanently deletes it for any other tests running in the same worker. Use Vitest's `vi.stubEnv()` to set values and `vi.unstubAllEnvs()` to safely restore the original environment.

src/commands/auth/auth.test.ts
```diff
@@ -270,6 +279,89 @@ describe('auth command', () => {
...
+        it('prints the stored token to stdout for pipe-safe consumption', async () => {
+            mockProbeApiToken.mockResolvedValueOnce({
+                token: 'tk_stored_1234567890',
+                metadata: STORED_METADATA,
+            })
+
+            const program = createProgram()
+            await program.parseAsync(['node', 'tw', 'auth', 'token', 'view'])
+
+            expect(writeSpy).toHaveBeenCalledWith('tk_stored_1234567890')
```
doistbot: [P2] This assertion only proves the token was written at some point. The whole point of `auth token view` is pipe-safe stdout, so this test should assert the complete stdout payload (and ideally that `console.log` was not used) rather than `toHaveBeenCalledWith(...)`. As written, it would still pass if a later change added extra stdout text before/after the token.

src/commands/auth/index.ts
scottlovegrove: Done in 086a493 — updated `src/lib/skills/content.ts` to describe `tw auth token view [--user <ref>]` (the previous `tw auth token --user <ref>` line was wrong both before and after this PR; the actual printer is the new `view` subcommand). Regenerated `skills/twist-cli/SKILL.md` via `npm run sync:skill`.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 086a493 — switched the env wiring to `vi.stubEnv(TOKEN_ENV_VAR, ...)` per setup with a single `vi.unstubAllEnvs()` in `afterEach`. A developer's exported `TWIST_API_TOKEN` is now restored after the suite, no cross-test leakage.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 086a493 — added a `stdoutPayload()` helper that joins every `process.stdout.write` call into a single string, then asserted the full payload (e.g. `expect(stdoutPayload()).toBe('tk_stored_1234567890')`) and that `consoleSpy` was not called. Any future preamble/trailer or stray `console.log` would now fail the pipe-safety contract.

rfgamaral (commented):
✅

PR #228, opened 2026-05-16
feat(auth): support pre-subcommand `tw --user <ref> <subcommand>`

scottlovegrove:
## Summary

cli-core's auth attachers (`attachLogoutCommand`, `attachStatusCommand`, `attachTokenViewCommand`) already declare a per-command `--user <ref>` flag, but commander has no root-level `--user` option, so the global form `tw --user 42 auth status` was rejected at parse time. This PR completes the plumbing started by the cli-core 0.16 adoption (#226).

- `src/index.ts` — warm the global-args cache via `getRequestedUserRef()` before mutating argv (so the original `--user` value is captured), then strip the pre-subcommand `--user` from `process.argv` via cli-core's `stripUserFlag` so commander sees a clean argv. Per-command `--user` (after the subcommand name) survives the strip and reaches commander unchanged.
- `src/lib/global-args.ts` — expose `getRequestedUserRef()` reading `store.get().user` from the existing lazy parser cache. The `user` field was already part of `TwGlobalArgs`; this just adds an accessor.
- `src/commands/auth/store-wrap.ts` *(new)* — `withUserRefAware(store, requestedRef)` adapter built with `Object.create(store)` and overriding only `active` / `clear` so future `TwistTokenStore` methods inherit automatically. The captured ref is substituted whenever the attachers call `active(ref)` / `clear(ref)` without an explicit ref; per-command `--user` wins because the explicit ref short-circuits the substitution. Pure adapter — does not reach into global state.
- `src/commands/auth/index.ts` — resolve `getRequestedUserRef()` once at registration and hand it to `withUserRefAware`; wire the wrapper to logout / status / token view (login does not accept `--user`).
- Skill docs (`src/lib/skills/content.ts` + regenerated `skills/twist-cli/SKILL.md`) — document the global form narrowed to the three auth subcommands that actually honour it.

**Single-user note:** until the keyring `TokenStore` + `migrateLegacyAuth` land (#2+#5 in the migration plan), the only valid `--user <ref>` is the lone stored account — any other ref surfaces `ACCOUNT_NOT_FOUND`. The plumbing being in place now means those follow-up PRs ship without extra wiring.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 3 new wiring tests at the user-visible boundary (global ref flows into `auth token view`, global ref flows into `auth status`, wrong global ref on `auth logout` surfaces `ACCOUNT_NOT_FOUND` despite cli-core's swallow-active-error contract); 614 total pass
- [x] `npm run build`
- [x] `tw --user 1 doctor --offline` — global flag accepted at root, doctor runs
- [x] `tw --user 1 auth --help` — strip works; commander does not reject `--user`
- [ ] Manual end-to-end after `tw auth login`:
  - `tw --user 1 auth status` → success
  - `tw auth status --user 1` → success (regression)
  - `tw --user 999 auth status` → `ACCOUNT_NOT_FOUND`
  - `tw --user 999 auth status --user 1` → success (per-command wins)
  - `tw --user 1 auth token view` → prints stored token


src/index.ts
```diff
@@ -112,6 +113,17 @@ for (const [name, [description]] of Object.entries(commands)) {
     }
 }
 
+// Commander has no root `--user` option. Trigger the global-args cache off
+// the original argv before stripping, so the value survives for downstream
+// consumers; then hand commander the stripped form.
+{
+    const originalArgs = process.argv.slice(2)
+    if (originalArgs.some((a) => a === '--user' || a.startsWith('--user='))) {
```
doistbot: [P3] This manual `originalArgs.some(...)` check reinvents flag parsing logic. You can avoid this duplicated scanning by calling `getRequestedUserRef()` unconditionally before `stripUserFlag` to safely and eagerly populate the cache using the existing parser.

src/commands/auth/auth.test.ts
```diff
@@ -374,6 +375,70 @@ describe('auth command', () => {
         })
     })
 
+    describe('global --user flag', () => {
+        // Tests simulate `src/index.ts`'s startup: mutate `process.argv` +
+        // `resetGlobalArgs()` to rebuild the parser cache, then hand
+        // commander the already-stripped argv (no `--user` token).
+        const STORED_METADATA = {
```
doistbot: [P3] This exact `STORED_METADATA` object is defined elsewhere in the file (inside `describe('token view subcommand')`). Lift it to the top-level `describe('auth command')` scope to reuse it across both test suites.

src/commands/auth/store-wrap.ts
```diff
@@ -0,0 +1,23 @@
...
+import { getRequestedUserRef } from '../../lib/global-args.js'
+
+// Bridge the global `tw --user <ref>` stripped by `src/index.ts` into
+// cli-core's attachers, which only see per-command `--user`. Explicit ref
+// (per-command) wins over the global form.
+export function withUserRefAware(store: TwistTokenStore): TwistTokenStore {
+    function effectiveRef(explicit: AccountRef | undefined): AccountRef | undefined {
+        return explicit ?? getRequestedUserRef()
+    }
+    const wrapped: TokenStore<TwistAccount> = {
```
doistbot: [P2] This adapter only changes `active()` and `clear()`, but it manually re-forwards every other `TwistTokenStore` method. Prefer starting from the original `store` object and overriding just those two methods so future additions to `TwistTokenStore` are inherited automatically instead of needing another wrapper update.

src/index.ts
```diff
@@ -112,6 +113,17 @@ for (const [name, [description]] of Object.entries(commands)) {
...
+        getRequestedUserRef()
+    }
+    process.argv = [process.argv[0], process.argv[1], ...stripUserFlag(originalArgs)]
```
doistbot: [P2] Stripping `--user` here for every invocation means subcommands that intentionally do **not** support `--user` now accept it silently. For example, `tw --user 1 auth login` and `tw --user 1 auth token abc` will parse even though those paths never use the ref. That weakens the command contract and makes a no-op flag look like targeted account selection. Please either scope the strip/forwarding to the auth subcommands that actually declare `--user`, or add an explicit validation path that rejects global `--user` for unsupported auth commands.

src/commands/auth/store-wrap.ts
```diff
@@ -0,0 +1,23 @@
+import type { AccountRef, TokenStore } from '@doist/cli-core/auth'
+import type { TwistAccount, TwistTokenStore } from '../../lib/auth-provider.js'
+import { getRequestedUserRef } from '../../lib/global-args.js'
+
+// Bridge the global `tw --user <ref>` stripped by `src/index.ts` into
+// cli-core's attachers, which only see per-command `--user`. Explicit ref
+// (per-command) wins over the global form.
+export function withUserRefAware(store: TwistTokenStore): TwistTokenStore {
+    function effectiveRef(explicit: AccountRef | undefined): AccountRef | undefined {
+        return explicit ?? getRequestedUserRef()
```
doistbot: [P2] This wrapper now depends on `src/index.ts` having primed the global-args cache before mutating `process.argv`. That hidden cross-layer coupling makes `withUserRefAware()` non-deterministic outside the main entrypoint and is why the tests have to simulate bootstrap state. A cleaner boundary is to resolve the requested ref once at the edge and pass it in, e.g. `withUserRefAware(store, requestedUserRef)`, so the adapter stays pure and reusable.

src/commands/auth/store-wrap.ts
```diff
@@ -0,0 +1,23 @@
...
+    }
+    const wrapped: TokenStore<TwistAccount> = {
+        active: (ref) => store.active(effectiveRef(ref)),
+        set: (account, token) => store.set(account, token),
+        clear: (ref) => store.clear(effectiveRef(ref)),
```
doistbot: [P2] Forwarding the implicit global ref into `clear()` makes `tw --user <ref> auth logout` probe the token store twice: cli-core already does an `active()` preflight, and `TwistTokenStore.clear(ref)` runs `resolveByRef()` again before clearing. The new auth tests explicitly note the double `probeApiToken()` path. Consider caching a successful implicit-ref lookup in the wrapper, or otherwise reusing the preflight result, so logout only hits the keyring once.

src/commands/auth/auth.test.ts
```diff
@@ -374,6 +375,70 @@ describe('auth command', () => {
         })
     })
 
+    describe('global --user flag', () => {
+        // Tests simulate `src/index.ts`'s startup: mutate `process.argv` +
```
doistbot, line 379: [P2] These tests manually recreate `src/index.ts` by mutating `process.argv` and passing Commander an already-stripped argv. That makes them brittle to bootstrap refactors and, more importantly, it means the actual root parsing path (`getRequestedUserRef()` warm-up + `stripUserFlag()`) is still untested. Please add one integration-style test around the real argv-normalization entrypoint, or extract that logic into a helper and test it directly.

src/commands/auth/auth.test.ts
```diff
@@ -374,6 +375,70 @@ describe('auth command', () => {
         })
     })
 
+    describe('global --user flag', () => {
```
doistbot, line 378: [P2] This new `global --user` suite covers `auth token view` and `auth logout`, but `registerAuthCommand()` in this PR also switched `auth status` to the ref-aware store. A regression in `tw --user <ref> auth status` would still pass all of these tests. Add at least one status case here, ideally including the per-command `--user` override precedence as well.

src/lib/skills/content.ts
```diff
@@ -27,6 +27,7 @@ tw auth token                    # Save API token manually (prompts securely; sc
 tw auth status                   # Verify authentication + show mode
 tw auth status --json            # Full status payload as JSON (--ndjson also supported)
 tw auth status --user <ref>      # Target a specific stored account (id, id:<n>, or display name)
+tw --user <ref> <subcommand>     # Same as the per-command --user, but accepted before any subcommand (e.g. tw --user 42 auth status)
```
doistbot: [P2] This new skill entry is broader than the implementation in this PR. `withUserRefAware()` is only wired into `auth status`, `auth logout`, and `auth token view`, so `tw --user <ref>` is not actually equivalent for arbitrary `<subcommand>` values. Per the repo’s skill-content guidance, examples should match real CLI behavior. Please narrow this line to the auth commands that honor the flag, or explicitly say other commands merely tolerate it without changing behavior.

doistbot (commented):
This PR successfully introduces global `--user` flag support for auth subcommands by stripping it before Commander evaluation and mapping the resolved reference through a new token store adapter. This is a great step toward fuller cli-core integration that cleanly sets up the infrastructure for upcoming multi-account features. There are a few areas that could use refinement, including tightening flag validation so unsupported commands do not silently accept the flag, decoupling the store adapter from global state, optimizing double-lookups during logout, expanding test coverage for the root parser and status command, and ensuring the skill documentation precisely matches the implemented scope.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F228&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3582b350-5169-11f1-8c95-b8ee5d5c8d2e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3582b350-5169-11f1-8c95-b8ee5d5c8d2e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/index.ts
scottlovegrove: Done in 153db4f — dropped the manual scan; `getRequestedUserRef()` is now called unconditionally so the parser populates the cache once with no duplicated `--user` detection.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 153db4f — lifted `STORED_METADATA` to the outer `describe('auth command')` scope; both subsuites now share the single definition.

src/commands/auth/store-wrap.ts
scottlovegrove: Done in 153db4f — rewrote as `Object.create(store)` and override only `active` / `clear`. Any future `TwistTokenStore` methods now resolve through the prototype chain instead of needing a manual passthrough here.

src/index.ts
scottlovegrove: Declining. To scope the strip per-subcommand we'd need to know which subcommand is being invoked before commander parses argv — chicken-and-egg. todoist-cli accepts the same universal-strip pattern in `src/index.ts`, and we already agreed off-thread to leave the related bare/empty-value usage guards out for the same minimal-scope reason. The flag is silently accepted by other commands but has no effect; happy to revisit if the silent-acceptance becomes a real user-confusion issue.

src/commands/auth/store-wrap.ts
scottlovegrove: Done in 153db4f — `withUserRefAware(store, requestedRef)` now takes the ref as a constructor argument. `auth/index.ts` resolves `getRequestedUserRef()` once at registration time and hands it in. The wrapper no longer imports from `global-args.js` and stays pure.

src/commands/auth/store-wrap.ts
scottlovegrove: Declining. The `clear()` path's second probe is a single keyring read whose worst case is a few ms; caching the preflight inside the wrapper would add per-call state and a lifecycle (invalidate when?) that's heavier than the win. Re-visit if profiling shows the keyring round-trip is hot.

src/commands/auth/auth.test.ts
scottlovegrove: Declining. Testing the real bootstrap path would require extracting the `src/index.ts` startup block into a helper purely to expose it to tests — and the strip + cache-warm logic is two cli-core calls (`getRequestedUserRef()` + `stripUserFlag(process.argv.slice(2))`), both already covered by cli-core's own tests. The brittleness risk is low because that block has no internal logic of its own; if cli-core's surface changes, types will catch it.

src/commands/auth/auth.test.ts
scottlovegrove: Done in 153db4f — added `threads `tw --user <ref> auth status` into the snapshot used by fetchLive` in the global flag block. A status-only regression in the `refAware` wiring would now fail this test alongside the existing token-view + logout cases.

src/lib/skills/content.ts
scottlovegrove: Done in 153db4f — narrowed the line to `tw --user <ref> auth <status|logout|token view>` and noted that other commands accept the flag but ignore it. Resynced `skills/twist-cli/SKILL.md`.

henningmu (commented):
Nice ✅

_(this might make my prior comment about cli-core moot 😅)_

PR #229, opened 2026-05-16
refactor(auth): add UserRecordStore adapter over current flat config (foundation for cli-core keyring store)

scottlovegrove:
## Summary

Foundation PR for adopting cli-core's `createKeyringTokenStore`. PR β will swap twist's custom `TwistTokenStore` + `lib/auth.ts` probe/save/clear over to cli-core's keyring-backed store; this PR adds the `UserRecordStore` adapter the swap will hand to cli-core.

- **New `src/lib/user-records.ts`** — `createTwistUserRecordStore()` returns a `UserRecordStore<TwistAccount>` that translates twist's current flat config fields (`authUserId` / `authUserName` / `authMode` / `authScope` / `token`) into cli-core's `UserRecord` list shape. Single-user today: `list()` returns at most one record; `upsert` and `remove` write or strip those flat fields; `getDefaultId` / `setDefaultId` map to `authUserId`.
- **No consumer wired** — `TwistTokenStore` and `auth.ts` probe/save/clear stay exactly as they are. Zero behaviour change.
- **No on-disk schema change** — the adapter writes the same flat fields twist already persists. Existing users see nothing different after this lands; the future swap (PR β) keeps the schema unchanged too via the same adapter.
- `token` is surfaced as `fallbackToken` (cli-core's name for the plaintext copy persisted when the OS keyring is unreachable at write time) — preserves twist's existing keyring-fallback behaviour.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 10 new adapter tests (list / upsert / remove / getDefaultId / setDefaultId, happy + edge cases including replace-not-merge for the token slot); 624 total pass
- [x] `npm run build`
- N/A manual smoke — no consumer wired, no user-facing behaviour change.

## Follow-up (PR β)

- Replace `createTwistTokenStore` with `createKeyringTokenStore({ serviceName: 'twist-cli', accountForUser: () => 'api-token', userRecords: createTwistUserRecordStore(), recordsLocation: getConfigPath() })`.
- Override `accountForUser` to the existing `'api-token'` keyring slot so tokens already in users' keychains stay readable — no keyring slot migration needed.
- Delete `lib/auth.ts` probe/save/clear; rewire callers (`config view`, `doctor`, `api`, etc.) to use `store.active()` / `store.set()` / `store.clear()` directly.
- Keep a thin `getApiToken()` shim with `TWIST_API_TOKEN` env precedence (cli-core's keyring store doesn't know about the env var).

### DIFF
--- a/src/lib/user-records.test.ts
+++ b/src/lib/user-records.test.ts
@@ -32,128 +32,60 @@ describe('createTwistUserRecordStore', () => {
         mocks.setConfig.mockReset().mockResolvedValue(undefined)
     })
 
-    describe('list()', () => {
-        it('returns an empty array when no authUserId is persisted', async () => {
-            mocks.getConfig.mockResolvedValue({ currentWorkspace: 7 })
+    it('round-trips a record through upsert + list', async () => {
+        mocks.getConfig.mockResolvedValue({ currentWorkspace: 7 })
+        const store = createTwistUserRecordStore()
+        const record = {
+            account: {
+                id: '42',
+                label: 'Ada',
+                authMode: 'read-write' as const,
+                authScope: 'user:read',
+            },
+            fallbackToken: 'tk_stored_1234567890',
+        }
 
-            expect(await createTwistUserRecordStore().list()).toEqual([])
-        })
+        await store.upsert(record)
+        // Simulate the just-written state so the subsequent list reads it back.
+        mocks.getConfig.mockResolvedValue(mocks.setConfig.mock.calls[0][0])
 
-        it('returns the single record reconstructed from flat config fields', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
-
-            expect(await createTwistUserRecordStore().list()).toEqual([
-                {
-                    account: {
-                        id: '42',
-                        label: 'Ada',
-                        authMode: 'read-write',
-                        authScope: 'user:read',
-                    },
-                    fallbackToken: 'tk_stored_1234567890',
-                },
-            ])
-        })
-
-        it('omits fallbackToken when no plaintext token is on disk', async () => {
-            mocks.getConfig.mockResolvedValue({ ...STORED_CONFIG, token: undefined })
-
-            const [record] = await createTwistUserRecordStore().list()
-
-            expect(record).toEqual({
-                account: {
-                    id: '42',
-                    label: 'Ada',
-                    authMode: 'read-write',
-                    authScope: 'user:read',
-                },
-            })
-            expect(record).not.toHaveProperty('fallbackToken')
-        })
+        expect(await store.list()).toEqual([record])
     })
 
-    describe('upsert(record)', () => {
-        it('writes account fields and replaces token when fallbackToken is set', async () => {
-            mocks.getConfig.mockResolvedValue({ currentWorkspace: 7 })
+    it('strips the token on upsert when fallbackToken is absent (cli-core replace-not-merge)', async () => {
+        // Without this, a stale plaintext token would survive a successful
+        // keyring-backed write and the runtime would prefer it over the
+        // fresh keyring value.
+        mocks.getConfig.mockResolvedValue(STORED_CONFIG)
 
-            await createTwistUserRecordStore().upsert({
-                account: {
-                    id: '42',
-                    label: 'Ada',
-                    authMode: 'read-write',
-                    authScope: 'user:read',
-                },
-                fallbackToken: 'tk_new_9876543210',
-            })
-
-            expect(mocks.setConfig).toHaveBeenCalledWith({
-                currentWorkspace: 7,
-                token: 'tk_new_9876543210',
+        await createTwistUserRecordStore().upsert({
+            account: {
+                id: '42',
+                label: 'Ada',
                 authMode: 'read-write',
                 authScope: 'user:read',
-                authUserId: 42,
-                authUserName: 'Ada',
-            })
+            },
         })
 
-        it('strips a stale token when fallbackToken is absent (replace-not-merge)', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
-
-            await createTwistUserRecordStore().upsert({
-                account: {
-                    id: '42',
-                    label: 'Ada',
-                    authMode: 'read-write',
-                    authScope: 'user:read',
-                },
-            })
-
-            const written = mocks.setConfig.mock.calls[0][0] as Config
-            expect(written).not.toHaveProperty('token')
-            expect(written.authUserId).toBe(42)
-        })
+        expect(mocks.setConfig.mock.calls[0][0]).not.toHaveProperty('token')
     })
 
-    describe('remove(id)', () => {
-        it('clears all auth fields when the id matches the stored account', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
+    it('remove is a no-op when the id does not match (cli-core contract)', async () => {
+        mocks.getConfig.mockResolvedValue(STORED_CONFIG)
 
-            await createTwistUserRecordStore().remove('42')
+        await createTwistUserRecordStore().remove('999')
 
-            expect(mocks.setConfig).toHaveBeenCalledWith({ currentWorkspace: 7 })
-        })
-
-        it('is a no-op when the id does not match the stored account', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
-
-            await createTwistUserRecordStore().remove('999')
-
-            expect(mocks.setConfig).toHaveBeenCalledWith(STORED_CONFIG)
-        })
+        expect(mocks.setConfig).toHaveBeenCalledWith(STORED_CONFIG)
     })
 
-    describe('default user pointer', () => {
-        it('getDefaultId returns the stored authUserId as a string', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
+    it('setDefaultId(null) clears authUserId without disturbing other fields', async () => {
+        mocks.getConfig.mockResolvedValue(STORED_CONFIG)
 
-            expect(await createTwistUserRecordStore().getDefaultId()).toBe('42')
-        })
+        await createTwistUserRecordStore().setDefaultId(null)
 
-        it('getDefaultId returns null when no record exists', async () => {
-            mocks.getConfig.mockResolvedValue({ currentWorkspace: 7 })
-
-            expect(await createTwistUserRecordStore().getDefaultId()).toBeNull()
-        })
-
-        it('setDefaultId(null) clears authUserId without disturbing other fields', async () => {
-            mocks.getConfig.mockResolvedValue(STORED_CONFIG)
-
-            await createTwistUserRecordStore().setDefaultId(null)
-
-            const written = mocks.setConfig.mock.calls[0][0] as Config
-            expect(written).not.toHaveProperty('authUserId')
-            expect(written.token).toBe('tk_stored_1234567890')
-            expect(written.currentWorkspace).toBe(7)
-        })
+        const written = mocks.setConfig.mock.calls[0][0] as Config
+        expect(written).not.toHaveProperty('authUserId')
+        expect(written.token).toBe('tk_stored_1234567890')
+        expect(written.currentWorkspace).toBe(7)
     })
 })
--- a/src/lib/user-records.test.ts
+++ b/src/lib/user-records.test.ts
@@ -2,7 +2,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 const mocks = vi.hoisted(() => ({
     getConfig: vi.fn(),
-    setConfig: vi.fn(),
+    updateConfig: vi.fn(),
 }))
 
 vi.mock('./config.js', async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


scottlovegrove:
@doistbot /review

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+        authMode: _m,
+        authScope: _s,
+        authUserId: _id,
+        authUserName: _n,
+        ...rest
+    } = config
+    return rest
+}
+
+function applySetDefault(config: Config, id: string | null): Config {
```
doistbot: [P2] Since `buildRecords` relies on `config.authUserId` to return the user, clearing only `authUserId` here effectively deletes the user from the store while leaving their `token` and PII (`authUserName`) orphaned in the config file.

Because this flat config structure cannot represent a "stored but not default" user, setting the default to `null` should either strip all auth fields (exactly like `applyRemove`) or be a no-op to avoid leaving stale, unreachable credentials on disk. (The corresponding test in `user-records.test.ts` should be updated to match).

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+ * `token` is surfaced as `fallbackToken` (cli-core's name for the plaintext
+ * copy persisted when the OS keyring is unreachable at write time).
+ */
+export function createTwistUserRecordStore(): UserRecordStore<TwistAccount> {
+    return {
+        async list() {
+            return buildRecords(await getConfig())
+        },
+        async upsert(record) {
+            await setConfig(applyUpsert(await getConfig(), record))
```
doistbot: [P2] This uses a read-modify-write cycle via `getConfig()` + `setConfig()`. `src/lib/config.ts` already distinguishes `updateConfig()` as the atomic path, and this pattern can lose concurrent writes to unrelated keys like `currentWorkspace`, `updateChannel`, or `userSettings` if another `tw` process updates the config between the read and write. Please switch these store mutations to an atomic config update helper before wiring this adapter into auth.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+        const { authUserId: _id, ...rest } = config
+        return rest
+    }
+    const numeric = Number(id)
+    if (!Number.isFinite(numeric)) return config
+    // Single-user adapter: cli-core's `KeyringTokenStore.setDefault` validates
+    // the id against `list()` first, so a non-matching ref reaches us only
+    // through programmer error — no-op rather than overwrite the stored
+    // identity.
+    if (config.authUserId !== undefined && config.authUserId !== numeric) {
```
doistbot: [P2] The guard here does not cover the “no stored identity” case. With an empty config, `setDefaultId('42')` falls through to line 100 and writes `authUserId: 42`, after which `list()` synthesizes a phantom account with blank label/scope. If this method is meant to no-op on programmer error, it should also return unchanged when `config.authUserId` is `undefined` instead of creating a new record.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+        },
+        async upsert(record) {
+            await setConfig(applyUpsert(await getConfig(), record))
+        },
+        async remove(id) {
+            await setConfig(applyRemove(await getConfig(), id))
+        },
+        async getDefaultId() {
+            const config = await getConfig()
+            return buildRecords(config).length > 0 && config.authUserId !== undefined
```
doistbot: [P3] The `buildRecords(config).length > 0` check duplicates the `authUserId` existence logic since `buildRecords()` only returns items when `config.authUserId !== undefined`. You can skip the function call entirely and simplify this return to: `return config.authUserId !== undefined ? String(config.authUserId) : null`.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+    const trimmed = config.token?.trim()
+    if (trimmed) record.fallbackToken = trimmed
+    return [record]
+}
+
+function applyUpsert(config: Config, record: UserRecord<TwistAccount>): Config {
+    const userId = Number(record.account.id)
+    const next: Config = {
+        ...config,
+        authUserId: Number.isFinite(userId) ? userId : config.authUserId,
```
doistbot: [P2] `applyUpsert()` is re-encoding the same `TwistAccount -> config` mapping that `createTwistTokenStore().set()` already does, and it has already drifted: a non-numeric `account.id` keeps the previous `authUserId` here, while the existing store clears it by passing `undefined` through to `saveApiToken()`. Please extract a shared serializer for these auth fields and reuse it in both places so the eventual store swap does not silently change this edge-case behavior.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+        authUserName: record.account.label,
+        authMode: record.account.authMode,
+        authScope: record.account.authScope,
+    }
+    // REPLACE semantics (per cli-core's contract): an absent `fallbackToken`
+    // means "no plaintext token", so the stored value must be cleared.
+    if (record.fallbackToken !== undefined) {
+        next.token = record.fallbackToken
```
doistbot: [P2] This writes `fallbackToken` verbatim, which duplicates the token-normalization logic in `auth.ts` instead of reusing it. The current auth flow trims tokens before persisting (`saveApiToken()` / `getConfigToken()`), so this adapter can now produce a different on-disk value for the same token. Please pull that normalization into a shared helper and use it here as well to keep both storage paths aligned.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+                : null
+        },
+        async setDefaultId(id) {
+            await setConfig(applySetDefault(await getConfig(), id))
+        },
+    }
+}
+
+function buildRecords(config: Config): UserRecord<TwistAccount>[] {
+    if (config.authUserId === undefined) return []
```
doistbot: [P1] Users who authenticated via `tw auth token` do not have an `authUserId` in their config. `buildRecords` currently ignores them entirely, meaning `list()` will return empty. When PR β swaps to `KeyringTokenStore`, `cli-core` won't find a user record and will fail to load their keyring token, effectively logging them out.

To preserve existing sessions, map an absent `authUserId` to `id: ''` (matching the current `loadStoredSnapshot` behaviour in `auth-provider.ts`). You can check `if (config.authUserId === undefined && config.authMode === undefined && config.token === undefined) return []` to detect a truly empty state.

You will also need to update `getDefaultId()` to read the ID from the yielded record instead of strictly checking `config.authUserId !== undefined`, and ensure `applyUpsert`/`applyRemove`/`applySetDefault` safely handle `id === ''` so `Number('')` doesn't inadvertently write `authUserId: 0` to disk or fail to clear the config.

src/lib/user-records.ts
```diff
@@ -0,0 +1,101 @@
...
+export function createTwistUserRecordStore(): UserRecordStore<TwistAccount> {
+    return {
+        async list() {
+            return buildRecords(await getConfig())
+        },
+        async upsert(record) {
+            await setConfig(applyUpsert(await getConfig(), record))
+        },
+        async remove(id) {
+            await setConfig(applyRemove(await getConfig(), id))
```
doistbot: [P2] `applyRemove()` returns the original `config` object when the id doesn't match, but `remove()` still calls `setConfig(...)`, which rewrites the config file anyway. Since this is a defined no-op path (and `setDefaultId()` has the same pattern below), short-circuit when the helper returns the same object to avoid an unnecessary disk write.

src/lib/user-records.test.ts
```diff
@@ -0,0 +1,91 @@
...
+const STORED_CONFIG: Config = {
+    token: 'tk_stored_1234567890',
+    authMode: 'read-write',
+    authScope: 'user:read',
+    authUserId: 42,
+    authUserName: 'Ada',
+    currentWorkspace: 7,
+}
+
+describe('createTwistUserRecordStore', () => {
```
doistbot, line 29: [P2] This suite still leaves key new adapter behavior unverified: there is no direct coverage for `getDefaultId()` and no test for the matching-id `remove()` path that actually clears the auth/token fields. Those are the branches the upcoming cli-core swap will rely on for active-user selection and logout cleanup, so a regression there would currently pass unnoticed.

doistbot (commented):
This PR introduces the `UserRecordStore` adapter to map the existing flat auth configuration into `cli-core`'s expected structure, laying a solid groundwork for the upcoming keyring migration. The approach safely prepares for the transition without altering the current on-disk schema. A few adjustments are needed before the final swap, specifically to preserve legacy token-only authentication sessions, ensure atomic configuration updates, and consolidate duplicated token normalization and state management logic.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F229&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A845994b0-51c0-11f1-8340-4e87255eb312%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A845994b0-51c0-11f1-8340-4e87255eb312&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/user-records.ts
scottlovegrove: Done — `setDefaultId(null)` is now a no-op (along with `setDefaultId(<unknown-id>)` per the sibling thread). The single-user adapter can't represent "stored but not default", so silently dropping the call beats either orphaning the token + label or rewriting the file with no semantic change. The test was updated to assert no `updateConfig` call for `null`, `<unknown>`, and `<matching>` invocations.

src/lib/user-records.ts
scottlovegrove: Done — every mutation now goes through `updateConfig({field: undefined, ...})`. cli-core's `updateConfig` does a single read-merge-write and only touches the keys you pass, so concurrent writes to unrelated keys (`currentWorkspace`, `updateChannel`, `userSettings`, etc.) can't be reverted by this adapter.

src/lib/user-records.ts
scottlovegrove: Done — `applySetDefault` now bails out before writing when `buildRecords(config).length === 0`, so `setDefaultId('42')` on an empty config no-ops instead of synthesising the phantom record. Covered by the consolidated `setDefaultId is a no-op` test.

src/lib/user-records.ts
scottlovegrove: Done — `getDefaultId` now reads the id off the synthesised record, which naturally folds the new legacy `id: ''` case into the same branch and avoids re-computing the `authUserId !== undefined` check.

src/lib/user-records.ts
scottlovegrove: Declining. The existing `createTwistTokenStore().set()` path goes away entirely in PR β — the adapter becomes the only writer of these fields. Extracting a shared serializer now would couple two paths that briefly coexist between α and β; after β lands there's nothing to share. The drift you noted (`Number.isFinite` guard) is intentionally tightened here (`> 0` now drops `0` and `NaN`); PR β will retire the old path.

src/lib/user-records.ts
scottlovegrove: Done — `applyUpsert` now trims `fallbackToken` before persisting and treats a blank result as "no token", matching `saveApiToken`'s normalisation. Same one-liner pattern.

src/lib/user-records.ts
scottlovegrove: Good catch — this is the P1 fix in this round. `buildRecords` now returns `[]` only when `authUserId`, `authMode`, and `token` are all undefined; otherwise it synthesises a record with `account.id = ''` matching the existing `loadStoredSnapshot` behaviour. `applyUpsert` writes `authUserId: undefined` for the `id === ''` (or non-numeric) case so the key drops from disk on serialisation rather than persisting `0`. `applyRemove` runs `buildRecords(config)` and compares against the resolved id, so removal of the synthesised record works too. New test (`synthesises a record with empty id for legacy token-only users`) pins this behaviour.

src/lib/user-records.ts
scottlovegrove: Done — both `remove` and `setDefaultId` short-circuit when the helper returns the original config object reference (cheap `===` check, no disk write on the no-op paths).

src/lib/user-records.test.ts
scottlovegrove: Done — added three new tests for the previously-uncovered paths: `synthesises a record with empty id for legacy token-only users` (covers the P1 fix and indirectly `getDefaultId` for the legacy case), `remove clears all auth fields when the id matches`, and `setDefaultId is a no-op for the single-user adapter` (covers all three branches: `null`, unknown id, matching id). The existing round-trip test covers `getDefaultId` for the happy path via `list()`.

PR #231, opened 2026-05-17
refactor(auth): swap custom TwistTokenStore for cli-core's keyring store

scottlovegrove:
## Summary

Replaces twist's hand-rolled `TwistTokenStore` + `lib/auth.ts` probe/save/clear with cli-core's `createKeyringTokenStore`, backed by the `UserRecordStore` adapter from #229. No on-disk schema change, no behavioural change for end users, no keyring slot migration.

**Net diff:** +324 / −1069. Most of the LOC is deletion — the new surface is a 5-line `createKeyringTokenStore({…})` call plus thin shims that preserve the existing `getApiToken` / `probeApiToken` / `getAuthMetadata` signatures.

## Changes

- **`lib/auth-provider.ts`** — `createTwistTokenStore()` is now `createKeyringTokenStore({ serviceName: 'twist-cli', accountForUser: () => 'api-token', userRecords: createTwistUserRecordStore(), recordsLocation: getConfigPath() })`. `accountForUser` is overridden to the existing `'api-token'` keyring slot so tokens already in users' keychains stay readable. The custom `~120 LOC` of `resolveByRef` / `matchesRef` / `loadStoredSnapshot` is deleted; cli-core owns all of it now. `createTwistAuthProvider` (OAuth flow) is untouched.
- **`lib/auth.ts`** — `getApiToken` / `probeApiToken` / `getAuthMetadata` are thin shims that check `TWIST_API_TOKEN` first then delegate to the store / record list. `saveApiToken` / `clearApiToken` / `cleanupAuthFallbackState` and friends are deleted. `probeApiToken`'s `source` field is derived by peeking at the `UserRecordStore.list()`: presence of `fallbackToken` indicates the keyring was unavailable at write time.
- **`commands/auth/token.ts`** — writes via `store.set({ id: '', label: '', authMode: 'unknown', authScope: '' }, token)` and reads `store.getLastStorageResult()` for the keyring-fallback warning. Token validation drops the arbitrary 10-char minimum in favour of non-empty-after-trim.
- **`commands/auth/helpers.ts`** — imports `TokenStorageResult` from `@doist/cli-core/auth` (cli-core's keyring store owns the shape).
- **`commands/auth/store-wrap.ts`** — adds an existence check via `store.list()` when substituting the global `--user` ref, so PR #228's `tw --user <wrong> auth logout` → `ACCOUNT_NOT_FOUND` behaviour survives the swap (cli-core's `KeyringTokenStore.clear` is a silent no-op on a non-matching ref).
- **Tests** — deletes the obsolete `createTwistTokenStore` custom-impl tests; rewrites `lib/auth.test.ts` and the token / logout / global-flag wiring tests against the new store-backed path.

## What this does NOT do

- No on-disk config schema change (still flat `authUserId` / `authUserName` / `authMode` / `authScope` / `token` fields)
- No keyring slot migration (`accountForUser: () => 'api-token'` keeps existing tokens reachable)
- No multi-user support (the `accountForUser` override + adapter's single-record shape both constrain it)
- No `migrateLegacyAuth` (only needed if we ever move to v2 `users[]`)

Multi-user remains a separate follow-up (PR γ).

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 589 pass (down from 624 after deleting ~35 obsolete custom-impl tests, plus 6 new shim tests + adjusted wiring cases)
- [x] `npm run build`
- [x] `tw auth --help` / `tw auth token --help` render the same surface
- [ ] Manual end-to-end after `tw auth login`:
  - `tw auth status` → success (existing keyring entry reads through cli-core)
  - `tw auth token view` → prints the bearer token
  - `tw auth token <new_token>` → overwrites (warning surfaces if keyring unavailable)
  - `tw auth logout` → clears (success line + warning surfaces if keyring unavailable)
- [ ] Manual fallback (keyring absent — temporarily move `node_modules/@napi-rs/keyring` aside): `tw auth token <token>` falls through to `config.json` with the "system credential manager unavailable" warning, then `tw auth status` reads back via the same fallback.

### DIFF
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -1,15 +1,35 @@
 import { Command } from 'commander'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
-// Mock the auth module
+// Mock the auth module (only the read-side shims are stubbed; the
+// write-side path now goes through `createTwistTokenStore` from
+// auth-provider.js, mocked below).
 vi.mock('../../lib/auth.js', async (importOriginal) => {
     const actual = await importOriginal<typeof import('../../lib/auth.js')>()
     return {
         ...actual,
-        saveApiToken: vi.fn(),
-        clearApiToken: vi.fn(),
         getAuthMetadata: vi.fn(),
         probeApiToken: vi.fn(),
+    }
+})
+
+// Mock the cli-core-backed token store so token / logout tests can drive
+// `set` / `clear` / `getLastStorageResult` / `getLastClearResult` directly.
+const storeMocks = vi.hoisted(() => ({
+    set: vi.fn(),
+    clear: vi.fn(),
+    active: vi.fn(),
+    list: vi.fn(),
+    setDefault: vi.fn(),
+    getLastStorageResult: vi.fn(),
+    getLastClearResult: vi.fn(),
+}))
+
+vi.mock('../../lib/auth-provider.js', async (importOriginal) => {
+    const actual = await importOriginal<typeof import('../../lib/auth-provider.js')>()
+    return {
+        ...actual,
+        createTwistTokenStore: () => storeMocks,
     }
 })
 
@@ -58,24 +78,14 @@ import { attachLoginCommand } from '@doist/cli-core/auth'
 import { TwistRequestError, type User } from '@doist/twist-sdk'
 import { createWrappedTwistClient } from '../../lib/api.js'
 import { type TwistAccount, type TwistTokenStore } from '../../lib/auth-provider.js'
-import {
-    clearApiToken,
-    getAuthMetadata,
-    NoTokenError,
-    probeApiToken,
-    saveApiToken,
-    TOKEN_ENV_VAR,
-} from '../../lib/auth.js'
+import { getAuthMetadata, TOKEN_ENV_VAR } from '../../lib/auth.js'
 import { resetGlobalArgs } from '../../lib/global-args.js'
 import { registerAuthCommand } from './index.js'
 import { attachTwistStatusCommand } from './status.js'
 
 const mockCreateInterface = vi.mocked(createInterface)
 
-const mockSaveApiToken = vi.mocked(saveApiToken)
-const mockClearApiToken = vi.mocked(clearApiToken)
 const mockGetAuthMetadata = vi.mocked(getAuthMetadata)
-const mockProbeApiToken = vi.mocked(probeApiToken)
 const mockCreateWrappedTwistClient = vi.mocked(createWrappedTwistClient)
 const mockAttachLoginCommand = vi.mocked(attachLoginCommand)
 
@@ -114,67 +124,49 @@ describe('auth command', () => {
         errorSpy.mockRestore()
     })
 
-    const STORED_METADATA = {
-        authMode: 'read-write' as const,
+    const STORED_ACCOUNT: TwistAccount = {
+        id: '1',
+        label: 'Test User',
+        authMode: 'read-write',
         authScope: 'user:read',
-        authUserId: 1,
-        authUserName: 'Test User',
-        source: 'secure-store' as const,
     }
 
+    const STORED_SNAPSHOT = { token: 'tk_stored_1234567890', account: STORED_ACCOUNT }
+    const STORED_RECORDS = [{ account: STORED_ACCOUNT, isDefault: true }]
+
     describe('token subcommand', () => {
-        it('successfully saves a token', async () => {
+        beforeEach(() => {
+            storeMocks.set.mockReset().mockResolvedValue(undefined)
+            storeMocks.getLastStorageResult.mockReset().mockReturnValue({ storage: 'secure-store' })
+        })
+
+        it('saves the token via store.set with an empty-id account, trims whitespace, and confirms', async () => {
             const program = createProgram()
-            const token = 'some_token_123456789'
 
-            // Mock successful token save
-            mockSaveApiToken.mockResolvedValue({ storage: 'secure-store' })
+            await program.parseAsync(['node', 'tw', 'auth', 'token', '  some_token_123456789  '])
 
-            await program.parseAsync(['node', 'tw', 'auth', 'token', token])
-
-            // Verify token was saved with unknown auth mode
-            expect(mockSaveApiToken).toHaveBeenCalledWith(token, { authMode: 'unknown' })
-
-            // Verify success message
+            expect(storeMocks.set).toHaveBeenCalledWith(
+                { id: '', label: '', authMode: 'unknown', authScope: '' },
+                'some_token_123456789',
+            )
             expect(consoleSpy).toHaveBeenCalledWith('✓', 'API token saved successfully!')
             expect(consoleSpy).toHaveBeenCalledWith(
                 'Token stored securely in the system credential manager',
             )
         })
 
-        it('handles saveApiToken errors', async () => {
+        it('lets store.set errors propagate unchanged', async () => {
+            storeMocks.set.mockRejectedValue(new Error('Permission denied'))
             const program = createProgram()
-            const token = 'some_token_123456789'
-
-            // Mock save failure
-            mockSaveApiToken.mockRejectedValue(new Error('Permission denied'))
 
             await expect(
-                program.parseAsync(['node', 'tw', 'auth', 'token', token]),
+                program.parseAsync(['node', 'tw', 'auth', 'token', 'some_token_123456789']),
             ).rejects.toThrow('Permission denied')
-
-            expect(mockSaveApiToken).toHaveBeenCalledWith(token, { authMode: 'unknown' })
         })
 
-        it('trims whitespace from token', async () => {
-            const program = createProgram()
-            const tokenWithWhitespace = '  some_token_123456789  '
-            const expectedToken = 'some_token_123456789'
-
-            mockSaveApiToken.mockResolvedValue({ storage: 'secure-store' })
-
-            await program.parseAsync(['node', 'tw', 'auth', 'token', tokenWithWhitespace])
-
-            expect(mockSaveApiToken).toHaveBeenCalledWith(expectedToken, { authMode: 'unknown' })
-        })
-
-        it('prompts interactively when no token argument given', async () => {
+        it('prompts interactively when no token argument is given', async () => {
             const originalIsTTY = process.stdin.isTTY
-            Object.defineProperty(process.stdin, 'isTTY', {
-                value: true,
-                configurable: true,
-            })
-            const program = createProgram()
+            Object.defineProperty(process.stdin, 'isTTY', { value: true, configurable: true })
             const mockRl = {
                 question: vi.fn((_prompt: string, cb: (answer: string) => void) => {
                     cb('interactive_token_456')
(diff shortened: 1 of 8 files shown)


src/commands/auth/store-wrap.ts
```diff
@@ -1,15 +1,34 @@
...
     requestedRef: AccountRef | undefined,
 ): TwistTokenStore {
+    async function effectiveRef(explicit: AccountRef | undefined): Promise<AccountRef | undefined> {
+        if (explicit !== undefined || requestedRef === undefined) {
+            return explicit
+        }
+        const records = await store.list()
+        if (
+            !records.some(
+                ({ account }) => account.id === requestedRef || account.label === requestedRef,
```
doistbot: [P1] This strict string equality check drops support for `id:<n>` prefixes and case-insensitive name matching, which is a regression from the previous custom store's `matchesRef` behavior. Consider importing `parseRef` from `../../lib/refs.js` to validate both `explicit` and `requestedRef` against the record list. You can then return the matched `account.id` so the underlying cli-core store can successfully process the command without needing to understand Twist's ref formats.

doistbot (commented):
This PR successfully simplifies the authentication layer by swapping out the custom token store for cli-core's keyring implementation, resulting in a great reduction in overall codebase size. Relying on the shared utility is an excellent step for maintainability, though a few edge cases around backward compatibility and abstraction boundaries need adjusting. Specifically, a few tweaks are necessary to ensure environment variables remain fully supported for status checks, to restore flexible user reference parsing like `id:` prefixes, and to ensure the legacy keyring contract and ref-normalization logic are properly covered by tests.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F231&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A2413fc20-51c6-11f1-8935-2638c2415c6d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A2413fc20-51c6-11f1-8935-2638c2415c6d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth-provider.ts
```diff
@@ -268,128 +270,11 @@ export function createTwistAuthProvider(): AuthProvider<TwistAccount> {
...
-            // Single-user store — already the default once `ref` matches.
-        },
-        getLastStorageResult() {
-            return lastStorageResult
-        },
-        getLastClearResult() {
-            return lastClearResult
-        },
-    }
+    return createKeyringTokenStore<TwistAccount>({
```
doistbot: [P1] Swapping `createTwistTokenStore()` to a bare `createKeyringTokenStore(...)` drops the old store's env-var path. The previous `active()` delegated to `probeApiToken()`, so `TWIST_API_TOKEN=... tw auth status` worked; this new store only knows about keyring + `userRecords`, and `attachTwistStatusCommand` still resolves auth through `store.active()`. The result is that env-backed auth now reports `NO_TOKEN` / unauthenticated. Please keep an env-aware wrapper around `active()` (or make status use the auth shim) so `TWIST_API_TOKEN` remains a first-class auth source.

src/lib/auth.ts
```diff
@@ -60,285 +38,89 @@ export type AuthProbeResult = {
...
-async function cleanupAuthFallbackState(
-    config: Config,
-    warningPrefix: string,
-): Promise<string | undefined> {
-    try {
-        await writeConfig(withoutAuthFallbackState(config))
-        return undefined
-    } catch (error) {
-        return buildConfigCleanupWarning(warningPrefix, error)
+    return {
```
doistbot, line 93: [P3] The logic to map an account object to `AuthMetadata` (`authMode`, `authScope`, `authUserId` with `toAuthUserId`, and `authUserName` fallbacks) is identical here and in `probeApiToken` (lines 92-96). Consider extracting a shared helper (e.g. `function toAuthMetadata(account, source)`) to keep this mapping single-sourced.

src/lib/auth.ts
```diff
@@ -60,285 +38,89 @@ export type AuthProbeResult = {
...
-    if (!token || token.trim().length < 10) {
-        throw new CliError('INVALID_TOKEN', 'Invalid token: Token must be at least 10 characters', [
-            'Run: tw auth login',
-            'Or set TWIST_API_TOKEN environment variable',
-        ])
-    }
-
-    const trimmedToken = token.trim()
-    const secureStore = getSecureStore()
+    const records = await createTwistUserRecordStore().list()
```
doistbot: [P2] `probeApiToken()` now reaches past `createTwistTokenStore()` into `createTwistUserRecordStore()` and infers provenance from `fallbackToken` / `records[0]` itself. That breaks the storage abstraction this PR is trying to introduce: `auth.ts` is now coupled to the current config-backed record format and single-user shape. Please move source/provenance derivation behind the token-store/provider seam so callers of `auth.ts` stay backend-agnostic.

src/commands/auth/store-wrap.ts
```diff
@@ -1,15 +1,34 @@
...
+            )
+        ) {
+            throw new CliError('ACCOUNT_NOT_FOUND', `No stored account matches "${requestedRef}".`)
+        }
+        return requestedRef
+    }
     return Object.assign(Object.create(store) as TwistTokenStore, {
-        active: (ref?: AccountRef) => store.active(ref ?? requestedRef),
-        clear: (ref?: AccountRef) => store.clear(ref ?? requestedRef),
+        active: async (ref?: AccountRef) => store.active(await effectiveRef(ref)),
```
doistbot: [P2] `active()` doesn't need the new `effectiveRef()` preflight. Once you substitute the captured global `--user` into `store.active(ref)`, cli-core can already do the normal ref-aware lookup itself; the added `store.list()` here just forces an extra config/record-store read before every `tw --user … auth status` / `auth token view`. Keep the existence check on `clear()` (where the silent no-op matters), but pass `ref ?? requestedRef` straight through for `active()`.

src/lib/auth-provider.test.ts
```diff
@@ -168,193 +152,3 @@ describe('createTwistAuthProvider', () => {
         })
     })
 })
-
-describe('createTwistTokenStore', () => {
```
doistbot, line 172: [P2] Removing the entire `createTwistTokenStore` suite leaves this PR without any test that the new factory still preserves the legacy keyring contract (`serviceName: 'twist-cli'`, `accountForUser() === 'api-token'`, and `recordsLocation === getConfigPath()`). Those parameters are the whole compatibility story for existing installs, so please keep a small unit test that mocks `createKeyringTokenStore` and asserts this wiring.

src/commands/auth/auth.test.ts
```diff
@@ -324,51 +277,49 @@ describe('auth command', () => {
...
             await expect(
-                program.parseAsync(['node', 'tw', 'auth', 'token', 'view']),
+                createProgram().parseAsync(['node', 'tw', 'auth', 'token', 'view']),
             ).rejects.toHaveProperty('code', 'NOT_AUTHENTICATED')
 
             expect(stdoutPayload()).toBe('')
         })
 
-        it('matches --user against the stored account by numeric id', async () => {
+        it('matches per-command --user against the stored account by id', async () => {
```
doistbot, line 299: [P2] This replacement coverage only exercises `--user 1`, but the deleted auth-store tests also protected `id:<n>` and case-insensitive label refs. Since the repo-wide ref contract still allows those forms, and the new global `--user` wrapper now has custom matching logic, add equivalent cases here (or next to `store-wrap`) so ref-normalization regressions are still caught.

src/commands/auth/store-wrap.ts
```diff
@@ -1,15 +1,34 @@
...
+import { CliError } from '../../lib/errors.js'
 
 // Bridge the global `tw --user <ref>` (stripped by `src/index.ts`) into
 // cli-core's attachers, which only see per-command `--user`. Explicit ref
-// passed by commander wins over the captured global ref.
+// passed by commander wins over the captured global ref. When a global ref
+// is substituted, validate existence here so the typed `ACCOUNT_NOT_FOUND`
+// surfaces — cli-core's `KeyringTokenStore.clear` is a silent no-op on a
+// non-matching ref, and `requireSnapshotForRef` only throws when the
+// per-command ref is set (which the global form deliberately strips).
```
scottlovegrove: If this is the reason we are doing this, does that mean it's a bug in cli-core?

src/commands/auth/store-wrap.ts
scottlovegrove: Done in ed0f7f7 — `createTwistTokenStore` now passes `matchAccount: matchTwistAccount` to `createKeyringTokenStore`, so cli-core itself handles `42` / `id:42` / case-insensitive labels via `parseRef`. The wrapper's existence check on `clear()` reuses the same matcher. New `matchTwistAccount` unit test pins all three forms.

src/lib/auth-provider.ts
scottlovegrove: Done in ed0f7f7 — `createTwistTokenStore` now wraps `active()` with env-var precedence. When no explicit `--user <ref>` is passed and `TWIST_API_TOKEN` is set, the wrap synthesises an env snapshot (account.id `''`, authMode `'unknown'`) and skips the keyring read entirely; with an explicit ref the env var is intentionally ignored so `tw --user 42 auth status` still targets the stored account. End-to-end coverage in the new `active() short-circuits to TWIST_API_TOKEN` test in `auth-provider.test.ts`.

src/lib/auth.ts
scottlovegrove: Done in ed0f7f7 — extracted a `toAccountFields(account)` helper used by both `probeApiToken` and `getAuthMetadata`.

src/lib/auth.ts
scottlovegrove: Done in ed0f7f7 — source detection moved into `getActiveTokenSource()` in `auth-provider.ts` (next to the store + record-store wiring). `lib/auth.ts` now calls that helper instead of reaching into the `UserRecordStore` directly, so the shim stays backend-agnostic.

src/commands/auth/store-wrap.ts
scottlovegrove: Done in ed0f7f7 — `active()` now passes `ref ?? requestedRef` straight through. cli-core's `KeyringTokenStore.active` returns `null` on a non-matching ref, and the attachers surface that via `onNotAuthenticated`. The existence pre-check stays on `clear()` where the silent no-op would otherwise mask `tw --user <wrong> auth logout`.

src/lib/auth-provider.test.ts
scottlovegrove: Done in ed0f7f7 — added a `createTwistTokenStore` wiring test in `auth-provider.test.ts` that mocks `createKeyringTokenStore` and asserts the four legacy-contract params: `serviceName: 'twist-cli'`, `accountForUser() === 'api-token'`, `recordsLocation === getConfigPath()`, and `matchAccount === matchTwistAccount`.

src/commands/auth/auth.test.ts
scottlovegrove: Done in ed0f7f7 — added a `matchTwistAccount` unit test that exercises numeric id, `id:<n>` prefix, and case-insensitive label refs. cli-core's `resolveTarget` (per-command `--user`) and the global wrapper's existence check both go through this matcher, so the three formats are covered uniformly.

src/commands/auth/store-wrap.ts
scottlovegrove: Good instinct — dug into cli-core and you're right. The silent no-op is a documented contract (`TokenStore.clear`: "No-op when `ref` doesn't match"), and cli-core's `attachLogoutCommand` already preflights for it via `requireSnapshotForRef` — but only when `cmd.user !== undefined`. Once we strip `--user` from argv pre-commander, cli-core's preflight is blind, and consumers have to compensate with a wrapper. Two CLIs (twist + todoist) now hand-roll the same ~50-LOC dance.

Raised upstream as [Doist/cli-core#30](https://github.com/Doist/cli-core/issues/30) — proposes a `getRequestedUserRef?: () => AccountRef | undefined` callback on each auth attacher's options so cli-core's preflight resolves the ref itself. Once that lands, `store-wrap.ts` goes away entirely from both consumer CLIs. Not blocking this PR.

nvignola:
✅

PR #232, opened 2026-05-17
feat(auth): schema v2 + multi-account migration foundation

scottlovegrove:
## Summary

Moves twist's on-disk auth state from flat scalar fields (`authUserId` / `authUserName` / `authMode` / `authScope` / `token`) into a versioned `users[]` array, transparently for existing installs.

Single-account installs look identical to the user after upgrade — same `tw auth status`, same workflows. The runtime is now multi-account capable internally (`tw auth login` twice creates two records); user-facing commands to manage stored accounts will follow separately.

## What's in

- **Schema v2** in `config.ts`: `CONFIG_VERSION = 2`, `StoredUser` type, `users[]` / `defaultUserId` / `config_version` fields + `validateConfigForDoctor` coverage
- **`UserRecordStore` rewrite** over `users[]` (true multi-record `list` / `upsert` / `remove`; `setDefaultId` persists; removal unpins the default)
- **Drop the `accountForUser='api-token'` override** — cli-core's default `user-${id}` slug is used; `migrateLegacyAuth` handles the in-place move from the old single-account slot
- **New `migrate-auth.ts`** wrapping cli-core's `migrateLegacyAuth` with twist callbacks (`hasMigrated` / `markMigrated` via `config_version` ≥ 2; `identifyAccount` via a raw `TwistApi` so migration sits outside the runtime auth graph; cleanup of v1 flat fields)
- **Postinstall hook** invokes migration silently; `createTwistTokenStore` also runs a memoised lazy `ensureMigrated()` so `--ignore-scripts` installs migrate on first CLI invocation
- **Offline-upgrade safety** — when `migrateLegacyAuth` can't complete (e.g. offline `identifyAccount`):
  - `active()` falls back to the legacy keyring slot + plaintext `config.token`, honouring `--user <ref>` against the synthetic account so it can't accidentally resolve to a different user
  - `set()` / `clear()` first discharge the legacy state on disk so manual `tw auth token <new>` / `tw auth logout` aren't shadowed by a stale v1 token on the next read
  - `getAuthMetadata` falls back to the v1 flat fields so a legacy `read-only` token still triggers `ensureWriteAllowed`'s local guard
  - `getActiveTokenSource` reports `'config-file'` when the legacy plaintext slot is still on disk, so doctor / config-view surfaces the security-relevant state
- **Shared helpers** in `auth-constants.ts` (`SECURE_STORE_SERVICE`, `LEGACY_KEYRING_ACCOUNT`) and `twist-account.ts` (`makeTwistAccount` / `toTwistAccount`) so the OAuth, migration, user-records, and legacy-fallback paths can't drift on account shape or storage slugs

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 622/622 pass
- [x] `npm run build`
- [x] Manual upgrade smoke: v1 install (token in `twist-cli` / `api-token` keyring slot, flat fields in `config.json`) → run `npm install` → `tw auth status` still authenticates; `config.json` now has `config_version: 2`, `users: [...]`, `defaultUserId: '<id>'`; legacy keyring slot empty; new `twist-cli` / `user-<id>` slot has the token
- [ ] Manual `--ignore-scripts` path: skip postinstall, run `tw auth status` — lazy fallback migrates on first invocation
- [x] Manual idempotence: re-run `tw auth status` — `hasMigrated()` short-circuits, no work happens
- [x] macOS keyring inspection: `security find-generic-password -s twist-cli -a user-<id>` ✓; `security find-generic-password -s twist-cli -a api-token` fails

### DIFF
--- a/src/lib/auth-provider.test.ts
+++ b/src/lib/auth-provider.test.ts
@@ -78,13 +78,7 @@ const STORED_ACCOUNT = {
 
 const json = (body: unknown, status = 200) => new Response(JSON.stringify(body), { status })
 
-/**
- * `createTwistTokenStore` memoises a module-level migration promise. To
- * exercise the cold-cache path in each test (and avoid shipping a
- * test-only reset export from the production module) we reset the
- * module registry and re-import. Mocks declared above re-apply on
- * re-import.
- */
+/** Reset the module-level migration memo for each test by re-importing. */
 async function loadCreateTwistTokenStore(): Promise<
     typeof import('./auth-provider.js').createTwistTokenStore
 > {
@@ -251,7 +245,7 @@ describe('createTwistTokenStore', () => {
         vi.unstubAllEnvs()
     })
 
-    it('passes twist-cli wiring to cli-core: serviceName, no accountForUser override (cli-core default user-${id} is used after γ1), the user-records adapter, the records location, and the parseRef-aware matcher', async () => {
+    it('passes twist-cli wiring to cli-core: serviceName, no accountForUser override (uses cli-core default `user-${id}`), records location, and the parseRef-aware matcher', async () => {
         const createTwistTokenStore = await loadCreateTwistTokenStore()
         createTwistTokenStore()
 
@@ -302,9 +296,7 @@ describe('createTwistTokenStore', () => {
         await store.set(STORED_ACCOUNT, 'tk')
         await store.setDefault('42')
 
-        // Migration runs exactly once even across mixed reads + writes — that's
-        // the "memoised" half of the contract. If it re-ran per call we'd
-        // double-charge the cold-cache penalty on every CLI invocation.
+        // Memoised: migration must run exactly once across mixed reads + writes.
         expect(migrateMocks.runMigrateLegacyAuth).toHaveBeenCalledTimes(1)
         expect(migrateMocks.runMigrateLegacyAuth).toHaveBeenCalledWith({ silent: true })
     })
@@ -376,11 +368,7 @@ describe('createTwistTokenStore', () => {
         expect(keyringMocks.createSecureStore).not.toHaveBeenCalled()
     })
 
-    it('falls back to legacy when runMigrateLegacyAuth itself rejects (the catch branch of ensureMigrated)', async () => {
-        // Exercises the `catch(() => null)` swallow inside `ensureMigrated`.
-        // Without this branch a thrown migration would propagate and the CLI
-        // would crash on first invocation post-upgrade for users with a
-        // half-broken keyring / disk.
+    it('falls back to legacy when runMigrateLegacyAuth rejects (catch branch of ensureMigrated)', async () => {
         migrateMocks.runMigrateLegacyAuth.mockRejectedValue(new Error('boom'))
         keyringMocks.secureStoreGetSecret.mockResolvedValue('tk_legacy_keyring')
         configMocks.getConfig.mockResolvedValue({
@@ -398,7 +386,7 @@ describe('createTwistTokenStore', () => {
         expect(keyringMocks.inner.active).not.toHaveBeenCalled()
     })
 
-    it('legacy snapshot synthesises account.id = "" for `tw auth token <token>` users with no authUserId (preserves the pre-γ1 single-user adapter behaviour)', async () => {
+    it('legacy snapshot synthesises account.id = "" for `tw auth token <token>` users with no authUserId on disk', async () => {
         migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
             status: 'skipped',
             reason: 'identify-failed',
@@ -417,9 +405,6 @@ describe('createTwistTokenStore', () => {
     })
 
     it('active(ref) ignores the legacy snapshot when an explicit --user targets a different account', async () => {
-        // Without this guard `tw <cmd> --user 999` while offline would silently
-        // authenticate as the default legacy user (id 42) instead of failing
-        // with the expected account-not-found outcome.
         migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
             status: 'skipped',
             reason: 'identify-failed',
@@ -463,9 +448,6 @@ describe('createTwistTokenStore', () => {
     })
 
     it('set() discharges legacy state on disk before writing v2, when migration is inconclusive', async () => {
-        // Without this, a manual `tw auth token <new>` during the offline
-        // window would land in v2 but the next invocation would still read
-        // the unchanged legacy token via the active() fallback path.
         migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
             status: 'skipped',
             reason: 'identify-failed',
--- a/src/lib/auth-provider.test.ts
+++ b/src/lib/auth-provider.test.ts
@@ -2,6 +2,10 @@ import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 
 vi.mock('./api.js', () => ({ createWrappedTwistClient: vi.fn() }))
 
+vi.mock('./migrate-auth.js', () => ({
+    runMigrateLegacyAuth: vi.fn().mockResolvedValue({ status: 'no-legacy-state' }),
+}))
+
 const keyringMocks = vi.hoisted(() => ({
     createKeyringTokenStore: vi.fn(),
     inner: {
@@ -35,6 +39,7 @@ vi.mock('./config.js', async (importOriginal) => {
 import { createWrappedTwistClient } from './api.js'
 import {
     AUTHORIZATION_URL,
+    __resetMigrationPromiseForTests,
     createTwistAuthProvider,
     createTwistTokenStore,
     matchTwistAccount,
@@ -197,18 +202,19 @@ describe('createTwistTokenStore', () => {
     beforeEach(() => {
         keyringMocks.createKeyringTokenStore.mockClear()
         keyringMocks.inner.active.mockReset()
+        __resetMigrationPromiseForTests()
     })
 
     afterEach(() => {
         vi.unstubAllEnvs()
     })
 
-    it('passes twist-cli wiring to cli-core: serviceName, the api-token slot, the user-records adapter, the records location, and the parseRef-aware matcher', () => {
+    it('passes twist-cli wiring to cli-core: serviceName, no accountForUser override (cli-core default user-${id} is used after γ1), the user-records adapter, the records location, and the parseRef-aware matcher', () => {
         createTwistTokenStore()
 
         const options = keyringMocks.createKeyringTokenStore.mock.calls[0][0]
         expect(options.serviceName).toBe('twist-cli')
-        expect(options.accountForUser('any-id')).toBe('api-token')
+        expect(options.accountForUser).toBeUndefined()
         expect(options.recordsLocation).toBe('/home/user/.config/twist-cli/config.json')
         expect(options.matchAccount).toBe(matchTwistAccount)
     })
--- a/src/lib/auth-provider.test.ts
+++ b/src/lib/auth-provider.test.ts
@@ -76,6 +76,19 @@ const STORED_ACCOUNT = {
     authScope: 'user:read',
 }
 
+const SKIPPED_RESULT = {
+    status: 'skipped',
+    reason: 'identify-failed',
+    detail: 'offline',
+} as const
+
+const LEGACY_CONFIG = {
+    authUserId: 42,
+    authUserName: 'Ada',
+    authMode: 'read-write' as const,
+    authScope: 'user:read',
+}
+
 const json = (body: unknown, status = 200) => new Response(JSON.stringify(body), { status })
 
 /** Reset the module-level migration memo for each test by re-importing. */
@@ -283,10 +296,6 @@ describe('createTwistTokenStore', () => {
 
     it('runs runMigrateLegacyAuth on the first store access and memoises across subsequent calls', async () => {
         keyringMocks.inner.active.mockResolvedValue(null)
-        keyringMocks.inner.list.mockResolvedValue([])
-        keyringMocks.inner.clear.mockResolvedValue(undefined)
-        keyringMocks.inner.set.mockResolvedValue(undefined)
-        keyringMocks.inner.setDefault.mockResolvedValue(undefined)
         const createTwistTokenStore = await loadCreateTwistTokenStore()
         const store = createTwistTokenStore()
 
@@ -302,18 +311,9 @@ describe('createTwistTokenStore', () => {
     })
 
     it('falls back to the legacy api-token keyring slot when migration is skipped (offline `identifyAccount`)', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
+        migrateMocks.runMigrateLegacyAuth.mockResolvedValue(SKIPPED_RESULT)
         keyringMocks.secureStoreGetSecret.mockResolvedValue('tk_legacy_keyring')
-        configMocks.getConfig.mockResolvedValue({
-            authUserId: 42,
-            authUserName: 'Ada',
-            authMode: 'read-write',
-            authScope: 'user:read',
-        })
+        configMocks.getConfig.mockResolvedValue(LEGACY_CONFIG)
         const createTwistTokenStore = await loadCreateTwistTokenStore()
 
         const snapshot = await createTwistTokenStore().active()
@@ -324,30 +324,18 @@ describe('createTwistTokenStore', () => {
         })
         expect(snapshot).toEqual({
             token: 'tk_legacy_keyring',
-            account: {
-                id: '42',
-                label: 'Ada',
-                authMode: 'read-write',
-                authScope: 'user:read',
-            },
+            account: { id: '42', label: 'Ada', authMode: 'read-write', authScope: 'user:read' },
         })
         // v2 path is not consulted when the legacy fallback succeeded.
         expect(keyringMocks.inner.active).not.toHaveBeenCalled()
     })
 
-    it('falls back to the legacy plaintext config.token when migration is skipped and the keyring is empty', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
-        keyringMocks.secureStoreGetSecret.mockResolvedValue(null)
+    it('falls back to plaintext config.token when migration is skipped and the keyring is empty', async () => {
+        migrateMocks.runMigrateLegacyAuth.mockResolvedValue(SKIPPED_RESULT)
         configMocks.getConfig.mockResolvedValue({
+            ...LEGACY_CONFIG,
             token: '  tk_legacy_plaintext  ',
-            authUserId: 42,
-            authUserName: 'Ada',
             authMode: 'read-only',
-            authScope: 'user:read',
         })
         const createTwistTokenStore = await loadCreateTwistTokenStore()
 
@@ -371,12 +359,7 @@ describe('createTwistTokenStore', () => {
     it('falls back to legacy when runMigrateLegacyAuth rejects (catch branch of ensureMigrated)', async () => {
         migrateMocks.runMigrateLegacyAuth.mockRejectedValue(new Error('boom'))
         keyringMocks.secureStoreGetSecret.mockResolvedValue('tk_legacy_keyring')
-        configMocks.getConfig.mockResolvedValue({
-            authUserId: 42,
-            authUserName: 'Ada',
-            authMode: 'read-write',
-            authScope: 'user:read',
-        })
+        configMocks.getConfig.mockResolvedValue(LEGACY_CONFIG)
         const createTwistTokenStore = await loadCreateTwistTokenStore()
 
         const snapshot = await createTwistTokenStore().active()
@@ -387,15 +370,8 @@ describe('createTwistTokenStore', () => {
     })
 
     it('legacy snapshot synthesises account.id = "" for `tw auth token <token>` users with no authUserId on disk', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
-        configMocks.getConfig.mockResolvedValue({
-            token: 'tk_token_only',
-            authMode: 'unknown',
-        })
+        migrateMocks.runMigrateLegacyAuth.mockResolvedValue(SKIPPED_RESULT)
+        configMocks.getConfig.mockResolvedValue({ token: 'tk_token_only', authMode: 'unknown' })
         const createTwistTokenStore = await loadCreateTwistTokenStore()
 
         const snapshot = await createTwistTokenStore().active()
@@ -404,60 +380,31 @@ describe('createTwistTokenStore', () => {
         expect(snapshot?.account.label).toBe('')
     })
 
-    it('active(ref) ignores the legacy snapshot when an explicit --user targets a different account', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
+    it('active(ref) returns the legacy snapshot when ref matches, falls through to v2 when it doesn’t', async () => {
+        migrateMocks.runMigrateLegacyAuth.mockResolvedValue(SKIPPED_RESULT)
         keyringMocks.secureStoreGetSecret.mockResolvedValue('tk_legacy_keyring')
-        configMocks.getConfig.mockResolvedValue({
-            authUserId: 42,
-            authUserName: 'Ada',
-            authMode: 'read-write',
-            authScope: 'user:read',
-        })
+        configMocks.getConfig.mockResolvedValue(LEGACY_CONFIG)
         keyringMocks.inner.active.mockResolvedValue(null)
         const createTwistTokenStore = await loadCreateTwistTokenStore()
+        const store = createTwistTokenStore()
 
-        const snapshot = await createTwistTokenStore().active('999')
+        const matched = await store.active('42')
+        expect(matched?.token).toBe('tk_legacy_keyring')
 
+        const mismatched = await store.active('999')
+        expect(mismatched).toBeNull()
         expect(keyringMocks.inner.active).toHaveBeenCalledWith('999')
-        expect(snapshot).toBeNull()
     })
 
-    it('active(ref) returns the legacy snapshot when the ref matches the legacy account', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
-        keyringMocks.secureStoreGetSecret.mockResolvedValue('tk_legacy_keyring')
-        configMocks.getConfig.mockResolvedValue({
-            authUserId: 42,
-            authUserName: 'Ada',
-            authMode: 'read-write',
-            authScope: 'user:read',
-        })
+    it('set() / clear() discharge legacy state on disk when migration is inconclusive', async () => {
+        migrateMocks.runMigrateLegacyAuth.mockResolvedValue(SKIPPED_RESULT)
         const createTwistTokenStore = await loadCreateTwistTokenStore()
+        const store = createTwistTokenStore()
 
-        const snapshot = await createTwistTokenStore().active('42')
+        await store.set(STORED_ACCOUNT, 'tk_new')
+        await store.clear('42')
 
-        expect(snapshot?.token).toBe('tk_legacy_keyring')
-        expect(keyringMocks.inner.active).not.toHaveBeenCalled()
-    })
-
-    it('set() discharges legacy state on disk before writing v2, when migration is inconclusive', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
-        const createTwistTokenStore = await loadCreateTwistTokenStore()
-
-        await createTwistTokenStore().set(STORED_ACCOUNT, 'tk_new')
-
-        expect(keyringMocks.secureStoreDeleteSecret).toHaveBeenCalledTimes(1)
+        expect(keyringMocks.secureStoreDeleteSecret).toHaveBeenCalledTimes(2)
         expect(configMocks.updateConfig).toHaveBeenCalledWith({
             token: undefined,
             authMode: undefined,
@@ -467,20 +414,6 @@ describe('createTwistTokenStore', () => {
             pendingSecureStoreClear: undefined,
         })
         expect(keyringMocks.inner.set).toHaveBeenCalledWith(STORED_ACCOUNT, 'tk_new')
-    })
-
-    it('clear() discharges legacy state when migration is inconclusive (so logout actually logs the user out)', async () => {
-        migrateMocks.runMigrateLegacyAuth.mockResolvedValue({
-            status: 'skipped',
-            reason: 'identify-failed',
-            detail: 'offline',
-        })
-        const createTwistTokenStore = await loadCreateTwistTokenStore()
-
-        await createTwistTokenStore().clear('42')
-
-        expect(keyringMocks.secureStoreDeleteSecret).toHaveBeenCalledTimes(1)
-        expect(configMocks.updateConfig).toHaveBeenCalled()
         expect(keyringMocks.inner.clear).toHaveBeenCalledWith('42')
     })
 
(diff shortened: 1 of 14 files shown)


src/lib/auth-provider.ts
```diff
@@ -286,18 +282,43 @@ export function matchTwistAccount(account: TwistAccount, ref: AccountRef): boole
...
+ * Memoised one-shot v1 → v2 migration trigger. The first read/write to the
+ * store on a cold process awaits it; later operations short-circuit on the
+ * resolved promise. Errors are swallowed (the cli-core helper itself runs in
+ * silent mode and leaves v1 state untouched on failure) so the CLI never
+ * fails to start because a migration attempt blew up.
+ */
+let migrationPromise: Promise<void> | undefined
+function ensureMigrated(): Promise<void> {
+    if (!migrationPromise) {
+        migrationPromise = runMigrateLegacyAuth({ silent: true })
```
doistbot: [P1] `runMigrateLegacyAuth()` can resolve with a non-success status like `{ status: 'skipped' }` (for example when `identifyAccount()` can't reach `getSessionUser()`, which your new tests model as `identify-failed/offline`). `ensureMigrated()` currently discards that result and proceeds to the v2-only `KeyringTokenStore` anyway. At that point we no longer read the legacy `api-token` slot or flat `config.token`, so an `--ignore-scripts` upgrade run while offline is treated as logged out even though the old credentials are still present. Please only switch to the new store after a real migration/already-migrated/no-legacy-state result, or keep the legacy read path available when migration is skipped.

src/lib/auth.ts
```diff
@@ -79,7 +79,10 @@ export async function probeApiToken(): Promise<AuthProbeResult> {
 /** Lightweight metadata read used by `tw auth status` once a token is confirmed. */
 export async function getAuthMetadata(): Promise<AuthMetadata> {
     if (process.env[TOKEN_ENV_VAR]) return { authMode: 'unknown', source: 'env' }
-    const [record] = await createTwistUserRecordStore().list()
+    const store = createTwistUserRecordStore()
+    const [records, defaultId] = await Promise.all([store.list(), store.getDefaultId()])
+    const record =
```
doistbot: [P2] The logic for resolving the active/default user record is exactly duplicated here and in `getActiveTokenSource` in `src/lib/auth-provider.ts`. Consider extracting a shared helper (e.g., `getDefaultUserRecord()`) to consolidate this logic.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+        identifyAccount: async (token) => {
+            const client = createWrappedTwistClient(token)
+            const user = await client.users.getSessionUser()
+            const config = await getConfig()
+            // Promote the v1 auth metadata onto the v2 account if it's there;
+            // legacy `tw auth token` users have no `authMode` / `authScope`,
+            // so we fall through to `'unknown'` / `''`, the same defaults
+            // `validateToken` would have written.
+            return {
```
doistbot: [P2] `identifyAccount()` is rebuilding `TwistAccount` with the same `id`/`label`/`authMode ?? 'unknown'`/`authScope ?? ''` defaults that `validateToken()` already applies in `auth-provider.ts`. Factor that into a shared helper so the migration path and the normal login path can't drift if the account shape or defaults change.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,66 @@
+import { type MigrateAuthResult, migrateLegacyAuth } from '@doist/cli-core/auth'
+import { createWrappedTwistClient } from './api.js'
```
doistbot: [P2] Importing `createWrappedTwistClient` here creates a new module cycle: `migrate-auth -> api -> auth -> auth-provider -> migrate-auth`. It works today because none of the bindings are touched during module initialization, but it couples the one-shot migration path to the runtime auth/token-store stack and makes future refactors brittle. Please identify the account with a minimal/raw Twist client helper in this module (or another lower-level module) so migration stays outside that graph.

src/lib/auth-provider.ts
```diff
@@ -325,6 +363,9 @@ export function createTwistTokenStore(): TwistTokenStore {
  */
 export async function getActiveTokenSource(): Promise<'env' | 'secure-store' | 'config-file'> {
     if (process.env[TOKEN_ENV_VAR]) return 'env'
-    const [record] = await createTwistUserRecordStore().list()
+    const store = createTwistUserRecordStore()
+    const [records, defaultId] = await Promise.all([store.list(), store.getDefaultId()])
```
doistbot: [P2] `store.list()` and `store.getDefaultId()` both independently `await getConfig()`, resulting in duplicate parallel reads of the config file. Consider calling `await getConfig()` once directly here and locating the active user via `config.users` and `config.defaultUserId` to avoid redundant I/O on CLI startup.

src/lib/auth-provider.test.ts
```diff
@@ -197,18 +202,19 @@ describe('createTwistTokenStore', () => {
...
         keyringMocks.inner.active.mockReset()
+        __resetMigrationPromiseForTests()
     })
 
     afterEach(() => {
         vi.unstubAllEnvs()
     })
 
-    it('passes twist-cli wiring to cli-core: serviceName, the api-token slot, the user-records adapter, the records location, and the parseRef-aware matcher', () => {
+    it('passes twist-cli wiring to cli-core: serviceName, no accountForUser override (cli-core default user-${id} is used after γ1), the user-records adapter, the records location, and the parseRef-aware matcher', () => {
```
doistbot: [P2] The lazy fallback migration (`ensureMigrated()`) is a critical path for users who install with `--ignore-scripts`, but there are no tests verifying that `runMigrateLegacyAuth` is actually triggered when store methods like `active()` or `set()` are called. Consider adding a test to ensure this memoized migration runs on first access.

src/lib/auth.test.ts
```diff
@@ -38,6 +42,7 @@ describe('auth shims over the cli-core keyring store', () => {
     beforeEach(() => {
         mocks.activeMock.mockReset()
         mocks.listMock.mockReset()
+        mocks.getDefaultIdMock.mockReset().mockResolvedValue(null)
```
doistbot: [P2] New `defaultId` fallback logic is introduced in `getAuthMetadata` (and `getActiveTokenSource`), but `getDefaultIdMock` is only ever set to return `null`. Consider adding tests to verify that the default record is correctly selected from the records array when `defaultId` is present.

src/lib/auth-provider.ts
```diff
@@ -286,18 +282,43 @@ export function matchTwistAccount(account: TwistAccount, ref: AccountRef): boole
...
+    if (!migrationPromise) {
+        migrationPromise = runMigrateLegacyAuth({ silent: true })
+            .then(() => undefined)
+            .catch(() => undefined)
+    }
+    return migrationPromise
+}
+
+/** Test-only — reset the memoised migration promise so subsequent reads re-run it. */
+export function __resetMigrationPromiseForTests(): void {
```
doistbot: [P2] This adds a test-only API to a production module, so it will be compiled into `dist/` and shipped with the CLI. Please keep the reset seam in test-only code instead (for example by re-importing the module per test or using a dedicated test helper) rather than exporting `__resetMigrationPromiseForTests()` from runtime code.

doistbot (commented):
This PR lays an excellent foundation for multi-account support by smoothly migrating the on-disk authentication state to a versioned schema and integrating legacy migration hooks. The thoughtful approach to lazy migration and extensive test suite are highly appreciated, as they guarantee a seamless upgrade path for existing users. There are a few areas to refine, specifically regarding the handling of skipped offline migrations to prevent unexpected logouts, resolving a module dependency cycle involving the Twist client, consolidating duplicated user resolution and config read logic, and expanding test coverage while keeping test-specific exports out of the production build.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F232&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ad6d6d9b0-51d3-11f1-8873-a32978d99843%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ad6d6d9b0-51d3-11f1-8873-a32978d99843&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth-provider.ts
scottlovegrove: Fixed in 7ee0969. `active()` now reads the migration result and, when it's not conclusive (`migrated` / `already-migrated` / `no-legacy-state`), falls back to `readLegacyTokenSnapshot()` which reads the legacy `api-token` keyring slot first and the plaintext `config.token` slot second. So an `--ignore-scripts` install that's offline still resolves the v1 credentials and keeps the user logged in until the next CLI invocation retries the migration. Covered by two new tests (`falls back to the legacy api-token keyring slot when migration is skipped` + the plaintext-config sibling).

src/lib/migrate-auth.ts
scottlovegrove: Fixed in 7ee0969. `migrate-auth.ts` no longer imports from `./api.js` — it uses a raw `new TwistApi(token)` from `@doist/twist-sdk` directly, which sits below the runtime auth/token-store stack. `identifyAccount` doesn't need the spinner wrapper anyway. `migrate-auth` is now: cli-core/auth + twist-sdk + config + twist-account + user-records — none of which transitively pull `auth-provider` at runtime.

src/lib/auth.test.ts
scottlovegrove: Added in 7ee0969. `src/lib/auth.test.ts` now has `getAuthMetadata picks the record matching defaultUserId, not the first one in users[]` (pinned default ≠ first) and `getAuthMetadata falls back to the first user when defaultUserId points at no stored record` (stale default). Also dropped the `getDefaultIdMock` indirection — the rewritten tests mock `getConfig` and exercise the real `getDefaultUserRecord` helper end-to-end.

src/lib/auth.ts
scottlovegrove: Fixed in 7ee0969. Extracted `getDefaultUserRecord(config)` in `src/lib/user-records.ts`. Both `getAuthMetadata` (`src/lib/auth.ts`) and `getActiveTokenSource` (`src/lib/auth-provider.ts`) now take an already-loaded `Config` and call the shared helper — single source of truth for default-resolution, single `getConfig()` per call site.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 7ee0969. `getActiveTokenSource` now does `getDefaultUserRecord(await getConfig())` — one config read, no `store.list()` + `store.getDefaultId()` round-trip. Same shared helper as `getAuthMetadata` (see the sibling thread).

src/lib/auth-provider.test.ts
scottlovegrove: Added in 7ee0969. New test `runs runMigrateLegacyAuth on the first store access and memoises across subsequent calls` in `src/lib/auth-provider.test.ts` exercises `active() → list() → clear() → set() → setDefault()` and asserts `runMigrateLegacyAuth` is invoked exactly once with `{ silent: true }`. The `--ignore-scripts` path is the whole point of that memo so it's now load-bearing.

src/lib/migrate-auth.ts
scottlovegrove: Fixed in 7ee0969. New `src/lib/twist-account.ts` exports `toTwistAccount(sessionUser, metadata)`. Both `validateToken` (in `auth-provider.ts`) and `identifyAccount` (in `migrate-auth.ts`) call it now, so the `authMode ?? 'unknown'` / `authScope ?? ''` defaults live in exactly one place. Put it in its own module to avoid re-introducing the cycle that the next thread is about.

src/lib/auth-provider.ts
scottlovegrove: Fixed in 7ee0969. `__resetMigrationPromiseForTests` is gone from `src/lib/auth-provider.ts`. The `createTwistTokenStore` describe block now resets the memo by calling `vi.resetModules()` + `await import('./auth-provider.js')` per test (same pattern as `src/postinstall.test.ts`). Verified with `grep __resetMigrationPromiseForTests src/ dist/` after build — no matches.

scottlovegrove:
@doistbot /review

src/lib/auth-provider.ts
```diff
@@ -286,18 +279,91 @@ export function matchTwistAccount(account: TwistAccount, ref: AccountRef): boole
...
+        token = await secureStore.getSecret()
+    } catch {
+        // Keyring unreachable — fall through to plaintext.
+    }
+    const config = await getConfig()
+    if (!token) token = config.token?.trim() || null
+    if (!token) return null
+    return {
+        token,
+        account: toTwistAccount(
```
doistbot: [P2] For legacy users who authenticated via `tw auth token <token>`, `config.authUserId` may be `undefined`. Passing `0` as a fallback causes `toTwistAccount` to synthesize `id: '0'`, whereas the previous legacy adapter correctly synthesized `id: ''` for this state. Consider constructing the `TwistAccount` object directly here to strictly preserve the legacy empty-ID behavior instead of forcing it through `toTwistAccount`.

src/lib/auth-provider.ts
```diff
@@ -313,8 +379,29 @@ export function createTwistTokenStore(): TwistTokenStore {
                     }
                 }
             }
+            const result = await ensureMigrated()
+            if (result === null || !migrationIsConclusive(result)) {
+                const legacy = await readLegacyTokenSnapshot()
+                if (legacy) return legacy
```
doistbot: [P2] `active()` ignores the explicitly requested `ref` when returning the legacy snapshot. If a user runs `tw <cmd> --user <ref>` while offline (so migration is skipped), they will be silently authenticated as the default legacy user instead of encountering the expected account-not-found outcome. Consider checking `ref` against the legacy account before returning it: `if (legacy && (!ref || matchTwistAccount(legacy.account, ref))) return legacy`.

src/lib/auth-provider.ts
```diff
@@ -313,8 +379,29 @@ export function createTwistTokenStore(): TwistTokenStore {
...
+            }
             return inner.active(ref)
         },
+        async set(account: TwistAccount, token: string) {
```
doistbot, line 399: [P1] These non-`active()` methods still operate only on the v2 backend when migration is skipped/inconclusive, but `active()` above keeps preferring the untouched legacy snapshot. That means `tw auth token <new>` or `tw auth logout` can report success while the next invocation still reads the old legacy credentials. Please either block writes/clears until migration is conclusive, or update/clear the legacy storage alongside the v2 store.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,66 @@
+import { type MigrateAuthResult, migrateLegacyAuth } from '@doist/cli-core/auth'
+import { TwistApi } from '@doist/twist-sdk'
+import type { TwistAccount } from './auth-provider.js'
+import { CONFIG_VERSION, getConfig, updateConfig } from './config.js'
+import { toTwistAccount } from './twist-account.js'
+import { createTwistUserRecordStore } from './user-records.js'
+
+const SERVICE_NAME = 'twist-cli'
```
doistbot: [P2] `serviceName` and the legacy keyring slot are now hard-coded here and again in `auth-provider.ts` (`'twist-cli'` / `'api-token'`). Extract these identifiers into one shared auth-storage constant module so the migrator, legacy fallback, and runtime keyring store can't drift if either value changes.

src/lib/user-records.ts
```diff
@@ -1,107 +1,94 @@
...
+ */
+export function getDefaultUserRecord(config: Config): UserRecord<TwistAccount> | null {
+    const users = config.users ?? []
+    if (users.length === 0) return null
+    const defaultId = config.defaultUserId
+    const user = (defaultId && users.find((u) => u.id === defaultId)) || users[0]
+    return toRecord(user)
+}
+
+function toRecord(user: StoredUser): UserRecord<TwistAccount> {
```
doistbot, line 60: [P2] `toRecord()` is still rebuilding `TwistAccount` with the same `authMode ?? 'unknown'` / `authScope ?? ''` defaults that `toTwistAccount()` was introduced to centralize. Reuse that helper here too (or extract a lower-level account factory) so stored-user reads can't drift from the OAuth and migration account-shaping logic.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,66 @@
...
+export async function runMigrateLegacyAuth(
+    options: { silent: boolean } = { silent: true },
+): Promise<MigrateAuthResult<TwistAccount>> {
+    return migrateLegacyAuth<TwistAccount>({
+        serviceName: SERVICE_NAME,
+        legacyAccount: LEGACY_KEYRING_ACCOUNT,
+        userRecords: createTwistUserRecordStore(),
+        hasMigrated: async () => {
+            const config = await getConfig()
+            return config.config_version === CONFIG_VERSION
```
doistbot: [P2] Tying the completion check to the exported `CONFIG_VERSION` constant assumes it will forever remain `2`. If `CONFIG_VERSION` is bumped to `3` in a future schema change, `config_version === CONFIG_VERSION` will evaluate to `false` for already-migrated v2 users, causing this v1→v2 migration hook to spuriously re-run. Consider checking `(config.config_version ?? 0) >= 2` here, and having `markMigrated` specifically write `2` (e.g. using a local `V2_SCHEMA_VERSION` constant) so this legacy migration is fully decoupled from future schema bumps.

src/lib/auth.ts
```diff
@@ -79,7 +79,7 @@ export async function probeApiToken(): Promise<AuthProbeResult> {
 /** Lightweight metadata read used by `tw auth status` once a token is confirmed. */
 export async function getAuthMetadata(): Promise<AuthMetadata> {
     if (process.env[TOKEN_ENV_VAR]) return { authMode: 'unknown', source: 'env' }
-    const [record] = await createTwistUserRecordStore().list()
+    const record = getDefaultUserRecord(await getConfig())
```
doistbot: [P1] `getAuthMetadata()` now reads only `users[]`, so on the `ensureMigrated() -> skipped` path it no longer describes the token that `createTwistTokenStore().active()` returned. `ensureWriteAllowed()` relies on this helper, which means a legacy read-only token is treated as `unknown` and mutating commands skip the local READ_ONLY guard on the first offline post-upgrade invocation. Please derive auth metadata from the same fallback snapshot that `active()` uses until migration is conclusive.

src/lib/migrate-auth.ts
```diff
@@ -0,0 +1,66 @@
...
+        },
+        markMigrated: async () => {
+            await updateConfig({ config_version: CONFIG_VERSION })
+        },
+        loadLegacyPlaintextToken: async () => {
+            const config = await getConfig()
+            return config.token?.trim() || null
+        },
+        identifyAccount: async (token) => {
+            const user = await new TwistApi(token).users.getSessionUser()
```
doistbot: [P3] `identifyAccount()` does the network call and the local `getConfig()` read sequentially even though they are independent. Starting them together with `Promise.all([new TwistApi(token).users.getSessionUser(), getConfig()])` would shave a config-read round trip off the first-run/postinstall migration path.

src/postinstall.ts
```diff
@@ -1,3 +1,8 @@
+import { runMigrateLegacyAuth } from './lib/migrate-auth.js'
 import { updateAllInstalledSkills } from './lib/skills/update-installed.js'
 
+// Each runs independently — failures must not break `npm install`. The lazy
+// fallback inside `createTwistTokenStore` covers users who installed with
+// `--ignore-scripts` or whose postinstall silently failed.
 updateAllInstalledSkills({ local: false }).catch(() => {})
+runMigrateLegacyAuth({ silent: true }).catch(() => {})
```
doistbot, line 7: [P2] This adds a second fire-and-forget side effect to postinstall, but `src/postinstall.test.ts` still only covers `updateAllInstalledSkills()`. Please add coverage that `runMigrateLegacyAuth({ silent: true })` is invoked on import and that a rejected migration promise is swallowed, otherwise this new install-time path can regress silently.

src/lib/config.ts
```diff
@@ -225,6 +267,61 @@ export function validateConfigForDoctor(config: Record<string, unknown>): string
         issues.push('update_channel must be one of: stable, pre-release')
     }
 
+    if (
```
doistbot, line 257: [P2] The new `validateConfigForDoctor()` branches for `config_version`, `defaultUserId`, and nested `users[]` records are untested — `src/lib/config.test.ts` still only exercises the older validation cases. Add a small accept/reject matrix for the v2 schema so malformed multi-account configs are caught by the doctor suite.

src/lib/auth-provider.ts
```diff
@@ -286,18 +279,91 @@ export function matchTwistAccount(account: TwistAccount, ref: AccountRef): boole
...
+ * fails to start because a migration attempt blew up; the runtime falls back
+ * to the legacy snapshot below when the result isn't conclusive.
+ *
+ * Module-private; tests reset it with `vi.resetModules()` + dynamic
+ * `import('./auth-provider.js')` so no test-only seam ships in dist/.
+ */
+let migrationPromise: Promise<MigrateAuthResult<TwistAccount> | null> | undefined
+function ensureMigrated(): Promise<MigrateAuthResult<TwistAccount> | null> {
+    if (!migrationPromise) {
+        migrationPromise = runMigrateLegacyAuth({ silent: true }).catch(() => null)
```
doistbot, line 349: [P2] The new lazy-migration tests only cover a resolved `{ status: 'skipped' }` result. There’s still no case for the `catch(() => null)` branch introduced here, even though that is the path for unexpected migration failures. Please add one test where `runMigrateLegacyAuth` rejects and assert `active()` still falls back to the legacy snapshot instead of consulting the v2 store.

src/lib/auth-provider.ts
```diff
@@ -325,6 +412,6 @@ export function createTwistTokenStore(): TwistTokenStore {
  */
 export async function getActiveTokenSource(): Promise<'env' | 'secure-store' | 'config-file'> {
     if (process.env[TOKEN_ENV_VAR]) return 'env'
-    const [record] = await createTwistUserRecordStore().list()
+    const record = getDefaultUserRecord(await getConfig())
```
doistbot: [P2] This helper now reads only `users[]`, so during a skipped/inconclusive migration it can report `secure-store` even when the token actually came from legacy plaintext `config.token`. That is the same read-side compatibility problem called out by the Service Production Readiness standard for persisted-data migrations: https://handbook.doist.com/doc/standard-service-production-readiness-qgNRng8TJx . Please preserve a v1 fallback here until the migration result is conclusive, otherwise `tw config view` / `tw doctor` can hide that credentials are still stored in plaintext.

doistbot (commented):
This PR lays a strong foundation for multi-account support by transitioning the auth state to a versioned array and introducing seamless migration hooks. The approach is well-structured and thoughtfully preserves the existing user experience during the transition. There are just a few edge cases to review regarding legacy fallback behavior during inconclusive migrations—where reads and writes can become desynchronized or misreport metadata—along with a few gaps in test coverage for the new validation branches and opportunities to consolidate hardcoded constants.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F232&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab64ac280-51d6-11f1-91a4-ed715acabbab%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab64ac280-51d6-11f1-91a4-ed715acabbab&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/auth.ts
scottlovegrove: Fixed in edb7f0f. `getAuthMetadata()` now falls back to the v1 flat fields (`token`, `authMode`, `authScope`, `authUserId`, `authUserName`) when no v2 record exists and any of those fields is still on disk — so a legacy `read-only` token reports its real mode and `ensureWriteAllowed` fires correctly during the post-upgrade offline window. Covered by the new `getAuthMetadata falls back to v1 flat fields when users[] is empty but legacy state is still on disk` test in `src/lib/auth.test.ts`.

src/lib/migrate-auth.ts
scottlovegrove: Fixed in edb7f0f. New `src/lib/auth-constants.ts` exports `SECURE_STORE_SERVICE = 'twist-cli'` and `LEGACY_KEYRING_ACCOUNT = 'api-token'`. `auth-provider.ts` (runtime store + legacy fallback + discharge helper) and `migrate-auth.ts` both import from there now, so the slot a writer parks a secret in can't drift from the slot a reader pulls it from.

src/lib/auth-provider.ts
scottlovegrove: Fixed in edb7f0f. `readLegacyTokenSnapshot` now constructs the account via `makeTwistAccount({ id: config.authUserId !== undefined ? String(config.authUserId) : '', label: config.authUserName ?? '', authMode, authScope })`. `tw auth token <token>` users (no `authUserId`) get `account.id === ''` exactly as the pre-γ1 single-user adapter did. Covered by the new `legacy snapshot synthesises account.id = "" for \`tw auth token <token>\` users` test in `src/lib/auth-provider.test.ts`.

src/lib/user-records.ts
scottlovegrove: Fixed in edb7f0f. Extracted a lower-level `makeTwistAccount({ id, label, authMode?, authScope? })` in `src/lib/twist-account.ts`; `toTwistAccount(sessionUser, metadata)` is now a thin wrapper that delegates to it. `toRecord` in `user-records.ts`, `readLegacyTokenSnapshot` in `auth-provider.ts`, and the OAuth/migration paths all go through the same factory now, so the `'unknown'` / `''` defaults live in exactly one place.

src/lib/migrate-auth.ts
scottlovegrove: Fixed in edb7f0f. `migrate-auth.ts` now defines a local `V2_SCHEMA_VERSION = 2`; `hasMigrated` reads `(config.config_version ?? 0) >= V2_SCHEMA_VERSION` and `markMigrated` writes exactly `2`. A future bump of the exported `CONFIG_VERSION` to 3 (or higher) won't make this v1→v2 helper spuriously re-run for already-migrated users. `migrate-auth.test.ts` updated to assert `config_version: 3` short-circuits `hasMigrated` to `true`.

src/lib/auth-provider.ts
scottlovegrove: Fixed in edb7f0f via the "clear legacy alongside v2" option. New `dischargeLegacyState()` helper deletes the legacy keyring slot + nulls out the v1 flat config fields, and it's invoked at the top of `set()` and `clear()` whenever `ensureMigrated()` returns a non-conclusive result. So `tw auth token <new>` and `tw auth logout` during the offline post-upgrade window genuinely take effect — the next CLI invocation reads the v2 record instead of falling through to the now-cleared legacy snapshot. Covered by three new tests: legacy discharge on `set()`, on `clear()`, and skip-the-discharge when migration was conclusive (no needless writes on the happy path).

src/lib/config.ts
scottlovegrove: Fixed in edb7f0f. `src/lib/config.test.ts` gains an accept/reject matrix for the v2 schema: well-formed v2 (`config_version`, `defaultUserId`, `users[]` with optional fields) accepted, plus rejection cases for non-integer `config_version`, non-string `defaultUserId`, non-array `users`, malformed `StoredUser` entries (wrong types, non-object), unknown `StoredUser` keys, invalid `authMode`, non-string `authScope` / `token`.

src/lib/migrate-auth.ts
scottlovegrove: Fixed in edb7f0f. `identifyAccount` now does `const [user, config] = await Promise.all([new TwistApi(token).users.getSessionUser(), getConfig()])`. New test `identifyAccount runs the API call and the local getConfig() concurrently` pauses the API resolution and asserts `getConfig` was already in flight before resolving.

src/lib/auth-provider.ts
scottlovegrove: Fixed in edb7f0f. New `falls back to legacy when runMigrateLegacyAuth itself rejects (the catch branch of ensureMigrated)` test in `src/lib/auth-provider.test.ts` — `runMigrateLegacyAuth.mockRejectedValue(new Error('boom'))` and asserts `active()` resolves the legacy snapshot instead of crashing or consulting the v2 store.

src/lib/auth-provider.ts
scottlovegrove: Fixed in edb7f0f. `active(ref)` now does `if (legacy && (ref === undefined || matchTwistAccount(legacy.account, ref))) return legacy` before delegating to `inner.active(ref)` — a ref that doesn't match the legacy snapshot falls through to the v2 store (which returns `null`, surfacing the account-not-found outcome). Two new tests in `src/lib/auth-provider.test.ts`: `active(ref) ignores the legacy snapshot when an explicit --user targets a different account` and `active(ref) returns the legacy snapshot when the ref matches the legacy account`.

src/postinstall.ts
scottlovegrove: Fixed in edb7f0f. `src/postinstall.test.ts` now mocks `./lib/migrate-auth.js` and adds two cases: `invokes runMigrateLegacyAuth({ silent: true }) on import` and `swallows errors from runMigrateLegacyAuth so a broken migration never blocks npm install`. Mirrors the existing `updateAllInstalledSkills` coverage pattern.

src/lib/auth-provider.ts
scottlovegrove: Fixed in edb7f0f. `getActiveTokenSource` now reads `getConfig()` once and applies the layered check: v2 record with `fallbackToken` → `'config-file'`, v2 record without → `'secure-store'`, no v2 record but legacy `config.token` on disk → `'config-file'` (the read-side-compatibility case the standard calls out). Means `tw config view` / `tw doctor` surface plaintext-on-disk during the skipped-migration window instead of misreporting it as keyring-backed.

src/lib/auth-provider.ts
```diff
@@ -286,22 +277,113 @@ export function matchTwistAccount(account: TwistAccount, ref: AccountRef): boole
...
         userRecords: createTwistUserRecordStore(),
         recordsLocation: getConfigPath(),
         matchAccount: matchTwistAccount,
     })
+    const maybeDischargeLegacy = async (): Promise<void> => {
+        const result = await ensureMigrated()
+        if (result === null || !migrationIsConclusive(result)) {
+            await dischargeLegacyState()
+        }
+    }
```
scottlovegrove: We prefer function over const.

src/lib/config.ts
```diff
@@ -47,20 +63,33 @@ export interface UserSettings {
     unarchiveNewThreads?: boolean
 }
 
+/**
+ * One row of the `users[]` array. `id` is the stringified numeric Twist user
+ * id. `token` is a plaintext fallback persisted only when the keyring is
+ * unavailable at write time.
+ */
+export interface StoredUser {
```
scottlovegrove: Types instead of interfaces.

scottlovegrove: Fixed in e4f5e9a — `StoredUser` is now a `type` alias. Left `UserSettings` and `Config` as interfaces since they predate this PR (out of scope here).

src/lib/auth-provider.ts
scottlovegrove: Fixed in e4f5e9a — `maybeDischargeLegacy` now declared with `async function` instead of arrow `const`.

craigcarlyle (commented):
✅

PR #233, opened 2026-05-17
feat(auth): tw account commands + multi-account config view

scottlovegrove:
## Summary

PR γ2 from the multi-account auth plan. Builds on #232 (schema v2 + migration foundation) to surface the multi-account capability.

- New `tw account` command tree:
  - `tw account list` — list stored accounts with default marker (`--json` emits `[{id, label, isDefault}, ...]`)
  - `tw account current` — show the active account; honours `TWIST_API_TOKEN` and reports source
  - `tw account use <ref>` — set the default stored account (id, `id:<n>`, or display name)
  - `tw account remove <ref>` — clear keyring + config entry; surfaces keyring-fallback warnings on stderr
- `use` / `remove` validate the ref against `store.list()` first so a non-matching ref surfaces as `ACCOUNT_NOT_FOUND` instead of cli-core's silent no-op
- `tw config view`:
  - new "Authenticated accounts" block listing `config.users[]` with default marker
  - `--json` now masks per-user `fallbackToken` plaintext (revealed with `--show-token`)
- Skill content regenerated (`tw skill` consumers will see the new commands after `tw skill update`)

No schema changes, no migration changes — pure addition on top of γ1.

## Test plan

- [x] `npm run type-check` — clean
- [x] `npm run lint:check` — clean
- [x] `npm test` — 39 files / 641 tests pass (13 new in `account.test.ts`, 4 new in `config.test.ts`)
- [x] `npm run build` + `npm run check:skill-sync` — clean
- [x] `node dist/index.js account --help` renders the new tree
- [x] Manual multi-account smoke (post-merge, against a real install):
  - `tw auth login` (account A) → `tw account list` shows one entry, marked default
  - `tw auth login` (account B, different OAuth grant) → `tw account list` shows two entries; A still default
  - `tw account use 'id:<B-id>'` → `tw account current` reports B
  - `tw --user 'id:<A-id>' auth status` → reports A (existing global-ref plumbing)
  - `tw account remove 'id:<A-id>'` → only B remains

### DIFF
--- a/src/commands/account/account.test.ts
+++ b/src/commands/account/account.test.ts
@@ -90,6 +90,14 @@ describe('account command', () => {
             expect(output).toContain('Default: id:1  Alan Grant')
         })
 
+        it('runs by default when no subcommand is given (tw account)', async () => {
+            seedStore([ACCOUNT_ALAN, 'default'])
+
+            await createProgram().parseAsync(['node', 'tw', 'account'])
+
+            expect(stdout()).toContain('Stored accounts (1)')
+        })
+
         it('reports the empty state when no accounts are stored', async () => {
             seedStore()
 
@@ -150,6 +158,19 @@ describe('account command', () => {
         it('emits {source:"legacy"} in --json mode for legacy snapshots', async () => {
             vi.stubEnv(TOKEN_ENV_VAR, '')
             storeMocks.active.mockResolvedValue(LEGACY_SNAPSHOT)
+
+            await createProgram().parseAsync(['node', 'tw', 'account', 'current', '--json'])
+
+            expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual({ source: 'legacy' })
+        })
+
+        it('reports a populated legacy snapshot (authUserId set) as legacy, not config', async () => {
+            // `readLegacyTokenSnapshot` populates id/label from v1 flat
+            // fields when present, so the empty-id check alone misses this
+            // case — `isLegacyAuthActive()` is the authoritative signal.
+            vi.stubEnv(TOKEN_ENV_VAR, '')
+            legacyMocks.isLegacyAuthActive.mockResolvedValue(true)
+            storeMocks.active.mockResolvedValue({ token: 'tk_legacy', account: ACCOUNT_ALAN })
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'current', '--json'])
 
--- a/src/commands/account/account.test.ts
+++ b/src/commands/account/account.test.ts
@@ -26,6 +26,7 @@ vi.mock('../../lib/auth-provider.js', async (importOriginal) => {
 
 vi.mock('chalk')
 
+import { ACCOUNT_ALAN, ACCOUNT_ELLIE } from '../../lib/__fixtures__/accounts.js'
 import { type TwistAccount } from '../../lib/auth-provider.js'
 import { TOKEN_ENV_VAR } from '../../lib/auth.js'
 import { registerAccountCommand } from './index.js'
@@ -37,18 +38,22 @@ function createProgram() {
     return program
 }
 
-const ACCOUNT_A: TwistAccount = {
-    id: '1',
-    label: 'Ada Lovelace',
-    authMode: 'read-write',
-    authScope: 'user:read',
+/** Seed `store.list()` and `store.setDefault/clear` resolvers in one call. */
+function seedStore(...records: Array<TwistAccount | [TwistAccount, 'default']>): void {
+    const list = records.map((spec) =>
+        Array.isArray(spec)
+            ? { account: spec[0], isDefault: true }
+            : { account: spec, isDefault: false },
+    )
+    storeMocks.list.mockResolvedValue(list)
+    storeMocks.setDefault.mockResolvedValue(undefined)
+    storeMocks.clear.mockResolvedValue(undefined)
+    storeMocks.getLastClearResult.mockReturnValue({ storage: 'secure-store' })
 }
 
-const ACCOUNT_B: TwistAccount = {
-    id: '2',
-    label: 'Bob Smith',
-    authMode: 'read-only',
-    authScope: 'user:read',
+const LEGACY_SNAPSHOT = {
+    token: 'tk_legacy',
+    account: { id: '', label: '', authMode: 'unknown' as const, authScope: '' },
 }
 
 describe('account command', () => {
@@ -72,24 +77,21 @@ describe('account command', () => {
 
     describe('list', () => {
         it('renders all stored accounts with the default marker', async () => {
-            storeMocks.list.mockResolvedValue([
-                { account: ACCOUNT_A, isDefault: true },
-                { account: ACCOUNT_B, isDefault: false },
-            ])
+            seedStore([ACCOUNT_ALAN, 'default'], ACCOUNT_ELLIE)
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'list'])
 
             const output = stdout()
             expect(output).toContain('Stored accounts (2)')
             expect(output).toContain('id:1')
-            expect(output).toContain('Ada Lovelace')
+            expect(output).toContain('Alan Grant')
             expect(output).toContain('id:2')
-            expect(output).toContain('Bob Smith')
-            expect(output).toContain('Default: id:1  Ada Lovelace')
+            expect(output).toContain('Ellie Sattler')
+            expect(output).toContain('Default: id:1  Alan Grant')
         })
 
         it('reports the empty state when no accounts are stored', async () => {
-            storeMocks.list.mockResolvedValue([])
+            seedStore()
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'list'])
 
@@ -99,16 +101,13 @@ describe('account command', () => {
         })
 
         it('emits a JSON envelope with id, label, isDefault', async () => {
-            storeMocks.list.mockResolvedValue([
-                { account: ACCOUNT_A, isDefault: true },
-                { account: ACCOUNT_B, isDefault: false },
-            ])
+            seedStore([ACCOUNT_ALAN, 'default'], ACCOUNT_ELLIE)
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'list', '--json'])
 
             expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual([
-                { id: '1', label: 'Ada Lovelace', isDefault: true },
-                { id: '2', label: 'Bob Smith', isDefault: false },
+                { id: '1', label: 'Alan Grant', isDefault: true },
+                { id: '2', label: 'Ellie Sattler', isDefault: false },
             ])
         })
     })
@@ -116,12 +115,12 @@ describe('account command', () => {
     describe('current', () => {
         it('renders the active account from store.active()', async () => {
             vi.stubEnv(TOKEN_ENV_VAR, '')
-            storeMocks.active.mockResolvedValue({ token: 'tk_abc', account: ACCOUNT_A })
+            storeMocks.active.mockResolvedValue({ token: 'tk_abc', account: ACCOUNT_ALAN })
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'current'])
 
             const output = stdout()
-            expect(output).toContain('Active account: id:1  Ada Lovelace')
+            expect(output).toContain('Active account: id:1  Alan Grant')
             expect(output).toContain('Mode:  read-write')
             expect(output).toContain('Scope: user:read')
         })
@@ -141,10 +140,7 @@ describe('account command', () => {
 
         it('renders a legacy-session notice when active() returns an empty-id snapshot', async () => {
             vi.stubEnv(TOKEN_ENV_VAR, '')
-            storeMocks.active.mockResolvedValue({
-                token: 'tk_legacy',
-                account: { id: '', label: '', authMode: 'unknown', authScope: '' },
-            })
+            storeMocks.active.mockResolvedValue(LEGACY_SNAPSHOT)
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'current'])
 
@@ -153,10 +149,7 @@ describe('account command', () => {
 
         it('emits {source:"legacy"} in --json mode for legacy snapshots', async () => {
             vi.stubEnv(TOKEN_ENV_VAR, '')
-            storeMocks.active.mockResolvedValue({
-                token: 'tk_legacy',
-                account: { id: '', label: '', authMode: 'unknown', authScope: '' },
-            })
+            storeMocks.active.mockResolvedValue(LEGACY_SNAPSHOT)
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'current', '--json'])
 
@@ -174,13 +167,13 @@ describe('account command', () => {
 
         it('emits a JSON envelope with the active account fields', async () => {
             vi.stubEnv(TOKEN_ENV_VAR, '')
-            storeMocks.active.mockResolvedValue({ token: 'tk_abc', account: ACCOUNT_A })
+            storeMocks.active.mockResolvedValue({ token: 'tk_abc', account: ACCOUNT_ALAN })
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'current', '--json'])
 
             expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual({
                 id: '1',
-                label: 'Ada Lovelace',
+                label: 'Alan Grant',
                 authMode: 'read-write',
                 authScope: 'user:read',
                 source: 'config',
@@ -190,11 +183,7 @@ describe('account command', () => {
 
     describe('use', () => {
         it('sets the default account by canonical id when the ref matches', async () => {
-            storeMocks.list.mockResolvedValue([
-                { account: ACCOUNT_A, isDefault: false },
-                { account: ACCOUNT_B, isDefault: true },
-            ])
-            storeMocks.setDefault.mockResolvedValue(undefined)
+            seedStore(ACCOUNT_ALAN, [ACCOUNT_ELLIE, 'default'])
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'use', '1'])
 
@@ -202,11 +191,11 @@ describe('account command', () => {
             expect(storeMocks.setDefault).toHaveBeenCalledWith('1')
             const output = stdout()
             expect(output).toContain('Default account set to')
-            expect(output).toContain('Ada Lovelace')
+            expect(output).toContain('Alan Grant')
         })
 
         it('rejects unknown refs with ACCOUNT_NOT_FOUND before touching the store', async () => {
-            storeMocks.list.mockResolvedValue([{ account: ACCOUNT_A, isDefault: true }])
+            seedStore([ACCOUNT_ALAN, 'default'])
 
             await expect(
                 createProgram().parseAsync(['node', 'tw', 'account', 'use', '999']),
@@ -216,41 +205,32 @@ describe('account command', () => {
         })
 
         it('matches refs by display name and resolves to the canonical id', async () => {
-            storeMocks.list.mockResolvedValue([
-                { account: ACCOUNT_A, isDefault: false },
-                { account: ACCOUNT_B, isDefault: true },
-            ])
-            storeMocks.setDefault.mockResolvedValue(undefined)
+            seedStore(ACCOUNT_ALAN, [ACCOUNT_ELLIE, 'default'])
 
-            await createProgram().parseAsync(['node', 'tw', 'account', 'use', 'ada lovelace'])
+            await createProgram().parseAsync(['node', 'tw', 'account', 'use', 'alan grant'])
 
             expect(storeMocks.setDefault).toHaveBeenCalledTimes(1)
             const output = stdout()
-            expect(output).toContain('Ada Lovelace')
-            expect(output).not.toContain('Bob Smith')
+            expect(output).toContain('Alan Grant')
+            expect(output).not.toContain('Ellie Sattler')
         })
     })
 
     describe('remove', () => {
         it('clears the account by canonical id and prints the removed label', async () => {
-            storeMocks.list.mockResolvedValue([
-                { account: ACCOUNT_A, isDefault: true },
-                { account: ACCOUNT_B, isDefault: false },
-            ])
-            storeMocks.clear.mockResolvedValue(undefined)
-            storeMocks.getLastClearResult.mockReturnValue({ storage: 'secure-store' })
+            seedStore([ACCOUNT_ALAN, 'default'], ACCOUNT_ELLIE)
 
-            await createProgram().parseAsync(['node', 'tw', 'account', 'remove', 'bob smith'])
+            await createProgram().parseAsync(['node', 'tw', 'account', 'remove', 'ellie sattler'])
 
             expect(storeMocks.clear).toHaveBeenCalledTimes(1)
             expect(storeMocks.clear).toHaveBeenCalledWith('2')
             const output = stdout()
             expect(output).toContain('Removed account')
-            expect(output).toContain('Bob Smith')
+            expect(output).toContain('Ellie Sattler')
         })
 
         it('rejects unknown refs with ACCOUNT_NOT_FOUND before clearing', async () => {
-            storeMocks.list.mockResolvedValue([{ account: ACCOUNT_A, isDefault: true }])
+            seedStore([ACCOUNT_ALAN, 'default'])
 
             await expect(
                 createProgram().parseAsync(['node', 'tw', 'account', 'remove', '999']),
@@ -260,8 +240,7 @@ describe('account command', () => {
         })
 
         it('surfaces keyring-fallback warnings on stderr', async () => {
-            storeMocks.list.mockResolvedValue([{ account: ACCOUNT_A, isDefault: true }])
-            storeMocks.clear.mockResolvedValue(undefined)
+            seedStore([ACCOUNT_ALAN, 'default'])
             storeMocks.getLastClearResult.mockReturnValue({
                 storage: 'config-file',
                 warning: 'system credential manager unavailable; local auth state cleared',
@@ -276,16 +255,14 @@ describe('account command', () => {
         })
 
         it('emits a JSON envelope and suppresses the plain confirmation', async () => {
-            storeMocks.list.mockResolvedValue([{ account: ACCOUNT_A, isDefault: true }])
-            storeMocks.clear.mockResolvedValue(undefined)
-            storeMocks.getLastClearResult.mockReturnValue({ storage: 'secure-store' })
+            seedStore([ACCOUNT_ALAN, 'default'])
 
             await createProgram().parseAsync(['node', 'tw', 'account', 'remove', '1', '--json'])
 
             expect(consoleSpy).toHaveBeenCalledTimes(1)
             expect(JSON.parse(consoleSpy.mock.calls[0][0] as string)).toEqual({
                 id: '1',
-                label: 'Ada Lovelace',
+                label: 'Alan Grant',
                 removed: true,
             })
         })
--- a/src/commands/account/account.test.ts
+++ b/src/commands/account/account.test.ts
@@ -11,11 +11,16 @@ const storeMocks = vi.hoisted(() => ({
     getLastClearResult: vi.fn(),
 }))
 
+const legacyMocks = vi.hoisted(() => ({
+    isLegacyAuthActive: vi.fn(),
+}))
+
 vi.mock('../../lib/auth-provider.js', async (importOriginal) => {
     const actual = await importOriginal<typeof import('../../lib/auth-provider.js')>()
     return {
         ...actual,
         createTwistTokenStore: () => storeMocks,
+        isLegacyAuthActive: legacyMocks.isLegacyAuthActive,
     }
 })
 
@@ -52,6 +57,7 @@ describe('account command', () => {
 
     beforeEach(() => {
         vi.clearAllMocks()
+        legacyMocks.isLegacyAuthActive.mockResolvedValue(false)
         consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
         errorSpy = vi.spyOn(console, 'error').mockImplementation(() => {})
     })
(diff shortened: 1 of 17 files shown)


src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    console.log(`  Mode:  ${account.authMode}`)
+    if (account.authScope) console.log(`  Scope: ${account.authScope}`)
+}
+
+async function useAccount(
+    ref: string,
+    options: ViewOptions,
+    store: TwistTokenStore,
+): Promise<void> {
+    const records = await store.list()
```
doistbot: [P2] This lookup is based on `store.list()`, but `createTwistTokenStore.list()` does not include the legacy single-user snapshot when auth migration is inconclusive, while `store.active()` does. In that state `tw account current` can still show the active legacy account, but `account use`/`remove` (and `account list`) will claim there is no matching stored account. Please either include the legacy snapshot in these paths or fail with a dedicated migration/unavailable error instead of returning `ACCOUNT_NOT_FOUND`.

src/commands/config/view.ts
```diff
@@ -81,6 +81,18 @@ function formatConfigView(
     lines.push(`${chalk.dim('Config file:')} ${getConfigPath()}${headerSuffix}`)
     lines.push('')
 
+    const users = config.users ?? []
+    if (users.length > 0) {
+        lines.push(chalk.bold(`Authenticated accounts (${users.length})`))
+        const defaultId = config.defaultUserId
+        for (const user of users) {
+            const isDefault = defaultId ? user.id === defaultId : false
```
doistbot: [P2] This only marks a row when `config.defaultUserId` is present and points at an existing user. The runtime selection logic falls back to the first stored user when `defaultUserId` is missing or stale, so `tw config view` can currently show no default marker even though one account will actually be used. Please derive the marker from the effective default record (`getDefaultUserRecord(config)` / first-user fallback) instead of `defaultUserId` alone.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    type TwistTokenStore,
+} from '../../lib/auth-provider.js'
+import { TOKEN_ENV_VAR } from '../../lib/auth.js'
+import { CliError } from '../../lib/errors.js'
+import { formatJson, formatNdjson } from '../../lib/output.js'
+import { logTokenStorageResult } from '../auth/helpers.js'
+
+type ViewOptions = { json?: boolean; ndjson?: boolean }
+
+function refLabel(account: TwistAccount): string {
```
doistbot: [P3] The `TwistAccount` type guarantees `id` and `label` are defined, making this defensive fallback unnecessary. Consider replacing this function with a formatting helper (e.g., `const formatAccount = (a) => `${chalk.dim(`id:${a.id}`)}  ${a.label}`). This can be reused across `listAccounts`, `currentAccount`, `useAccount`, and `removeAccount` to standardize output and fix the missing `chalk.dim` styling in the `use` and `remove` command text.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    return account.id ? `id:${account.id}` : account.label || '(unknown)'
+}
+
+function findAccount(
```
doistbot: [P3] This function duplicates the account validation and `ACCOUNT_NOT_FOUND` error throwing logic found in `src/commands/auth/store-wrap.ts`. Consider extracting this into `src/lib/auth-provider.ts` alongside `matchTwistAccount` so both command groups can share it. Reusing it there would also give `tw auth logout <ref>` the helpful `'Run: tw account list'` error hint.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    matchTwistAccount,
+    type TwistAccount,
+    type TwistTokenStore,
+} from '../../lib/auth-provider.js'
+import { TOKEN_ENV_VAR } from '../../lib/auth.js'
+import { CliError } from '../../lib/errors.js'
+import { formatJson, formatNdjson } from '../../lib/output.js'
+import { logTokenStorageResult } from '../auth/helpers.js'
+
+type ViewOptions = { json?: boolean; ndjson?: boolean }
```
doistbot: [P2] This adds a new ad-hoc output-options type even though `src/lib/options.ts` already defines the shared `ViewOptions` / `MutationOptions` shapes used across commands. Reusing those shared types here (or extending them once there for this command’s needs) would avoid a second source of truth for `json`/`ndjson` handling and keep `account use/remove` aligned with the rest of the CLI.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+    const match = records.find(({ account }) => matchTwistAccount(account, ref))
+    if (!match) {
+        throw new CliError('ACCOUNT_NOT_FOUND', `No stored account matches "${ref}".`, [
+            'Run: tw account list',
+        ])
+    }
+    return match.account
+}
+
+async function listAccounts(options: ViewOptions, store: TwistTokenStore): Promise<void> {
```
doistbot: [P2] According to `AGENTS.md`, commands with multiple subcommands should use a folder-based structure where handler functions are placed in sibling files (e.g., `list.ts`, `current.ts`, `use.ts`, `remove.ts`), and each exports one async action function. Please extract these handlers (and their tests) into separate files, leaving `index.ts` strictly for wiring Commander subcommands.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+}
+
+async function useAccount(
+    ref: string,
+    options: ViewOptions,
+    store: TwistTokenStore,
+): Promise<void> {
+    const records = await store.list()
+    const account = findAccount(records, ref)
+    await store.setDefault(ref)
```
doistbot: [P2] `findAccount()` already resolves the canonical account here, but `setDefault(ref)` re-runs matching in the store using the original user string. That means the account you validated is not guaranteed to be the one that gets mutated if matching semantics ever diverge (for example duplicate labels). Pass `account.id` into the store call instead so resolution stays single-sourced. The same issue applies to `store.clear(ref)` below.

src/commands/account/index.ts
```diff
@@ -0,0 +1,209 @@
...
+        return
+    }
+
+    const snapshot = await store.active()
+    if (!snapshot) {
+        throw new CliError('NO_TOKEN', 'No stored account is currently active.', [
+            'Run: tw auth login',
+        ])
+    }
+    const { account } = snapshot
```
doistbot: [P2] This assumes every non-env `store.active()` snapshot corresponds to a real stored account, but the auth-provider intentionally synthesizes legacy fallback snapshots with empty `id`/`label` when only a v1 token exists. In that case `tw account current` will emit an "active account" with blank fields. Please detect account-less snapshots and report them as source-only / no stored account, or resolve the live session user before rendering account details.

src/commands/account/account.test.ts
```diff
@@ -0,0 +1,266 @@
...
+
+            await createProgram().parseAsync(['node', 'tw', 'account', 'current'])
+
+            const output = consoleSpy.mock.calls.map((c: unknown[]) => c[0]).join('\n')
+            expect(output).toContain('Active account: id:1  Ada Lovelace')
+            expect(output).toContain('Mode:  read-write')
+            expect(output).toContain('Scope: user:read')
+        })
+
+        it('reports env-sourced tokens without inventing an account', async () => {
```
doistbot: [P2] The env-token path is only exercised in pretty output here. `account current` has separate `--json`/`--ndjson` branches, and those are the more important contract for scripting because they must emit only the source payload and skip `store.active()`. Please add a machine-output case for the env-sourced path so regressions there are caught.

src/commands/account/account.test.ts
```diff
@@ -0,0 +1,266 @@
...
+        it('matches refs by display name (case-insensitive)', async () => {
+            storeMocks.list.mockResolvedValue([
+                { account: ACCOUNT_A, isDefault: false },
+                { account: ACCOUNT_B, isDefault: true },
+            ])
+            storeMocks.setDefault.mockResolvedValue(undefined)
+
+            await createProgram().parseAsync(['node', 'tw', 'account', 'use', 'ada lovelace'])
+
+            expect(storeMocks.setDefault).toHaveBeenCalledWith('ada lovelace')
```
doistbot: [P2] This assertion is coupled to the implementation detail that the handler forwards the raw display-name ref into `store.setDefault()`. A harmless ref-normalization change (for example, resolving to the matched account id first) would preserve behavior but break this test. Prefer asserting the observable outcome here—e.g. successful selection of Ada plus a single store call—rather than the exact argument string.

src/lib/skills/content.ts
```diff
@@ -33,6 +33,12 @@ tw auth logout --json            # Emits \`{"ok": true}\` (--ndjson is silent)
 tw auth logout --user <ref>      # Target a specific stored account; mismatched ref errors with ACCOUNT_NOT_FOUND
 tw auth token view               # Print the saved token to stdout (pipe-safe; refuses if TWIST_API_TOKEN is set)
 tw auth token view --user <ref>  # Print the saved token for a specific stored account
+tw account list                  # List stored CLI accounts (default marked)
```
doistbot: [P2] The skill source-of-truth is still incomplete for the new `account` tree. The command implementation supports `--json`/`--ndjson` on every subcommand, but this block only documents `list --json` and omits the machine-output variants for `current`, `use`, and `remove` (plus the implicit default `tw account` behavior). Repo guidance requires `SKILL_CONTENT` to stay aligned with added commands/flags, so please document the full surface here and regenerate `skills/twist-cli/SKILL.md`.

doistbot (commented):
This PR successfully introduces the new `tw account` command tree to surface multi-account capabilities and updates the configuration view to display authenticated accounts. These additions provide a solid foundation for account management and smoothly build upon the recent schema migration work. There are a few areas to refine, particularly around extracting command handlers into separate files per the project architecture, reusing shared output options and validation logic, ensuring consistent behavior with legacy snapshots, and updating the skill content to fully reflect the new flags.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F233&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aa93dbb90-51e2-11f1-801c-bb9fa7845718%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aa93dbb90-51e2-11f1-801c-bb9fa7845718&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/account/account.test.ts
```diff
@@ -0,0 +1,313 @@
...
+            })
+        })
+    })
+
+    describe('use', () => {
+        it('sets the default account by canonical id when the ref matches', async () => {
+            storeMocks.list.mockResolvedValue([
+                { account: ACCOUNT_A, isDefault: false },
+                { account: ACCOUNT_B, isDefault: true },
+            ])
```
scottlovegrove: These test setups are repeated a lot, we should extract a helper for this

src/commands/account/index.ts
```diff
@@ -0,0 +1,49 @@
...
+import { currentAccount } from './current.js'
+import { listAccounts } from './list.js'
+import { removeAccount } from './remove.js'
+import { useAccount } from './use.js'
+
+function withJsonFlags(cmd: Command): Command {
+    return cmd
+        .option('--json', 'Output as JSON')
+        .option('--ndjson', 'Output as newline-delimited JSON')
+}
```
scottlovegrove, line 13: Are these flags not being picked up by cli-core?

src/commands/config/config.test.ts
```diff
@@ -273,6 +273,98 @@ describe('config view', () => {
         consoleSpy.mockRestore()
     })
 
+    it('renders an "Authenticated accounts" block from config.users with default marker', async () => {
+        presentConfig({
+            users: [
+                { id: '1', name: 'Ada Lovelace', authMode: 'read-write' },
```
scottlovegrove: I wonder if we should have fixtures for users for code re-use

src/commands/account/index.ts
scottlovegrove: Both, actually — `--json`/`--ndjson` still need to be registered with commander (else commander errors on unknown flag), but cli-core exports `emitView(view, payload, humanLines)` which does the `if (json) / elif (ndjson) / else humanLines()` dispatch and keeps `humanLines` lazy (so chalk strings aren't built in machine mode). I was reimplementing that inline. Fixed in 7d61f4f — `current.ts` / `use.ts` / `remove.ts` now call `emitView` from `@doist/cli-core`. `list.ts` keeps direct `formatJson/formatNdjson` because the payload is an array and `emitView` wraps single records.

src/commands/account/account.test.ts
scottlovegrove: Done in 7d61f4f — added `seedStore(...records)` that takes a list of `TwistAccount | [TwistAccount, 'default']` and seeds `store.list()` + the mutation mocks (`setDefault` / `clear` / `getLastClearResult`) in one call. Most tests now read as e.g. `seedStore([ACCOUNT_ALAN, 'default'], ACCOUNT_ELLIE)`.

src/commands/config/config.test.ts
scottlovegrove: Agreed — added `src/lib/__fixtures__/accounts.ts` in 7d61f4f, exporting `STORED_ALAN` / `STORED_ELLIE` (and `ACCOUNT_ALAN` / `ACCOUNT_ELLIE` derived by spreading the STORED versions + adding the account-only fields like `label` and `authScope`). Both `config.test.ts` and `account.test.ts` import from there so the names/ids are sourced once.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. Added `isLegacyAuthActive()` to auth-provider.ts; `account list/use/remove` now call `assertV2Available()` first and throw a dedicated `AUTH_MIGRATION_PENDING` envelope (with hints) when the legacy v1 token is still active. `account current` still surfaces the legacy session (see #8 below) since that is its job.

src/commands/config/view.ts
scottlovegrove: Fixed in 85097bc. `tw config view` now derives the default marker from `getDefaultUserRecord(config)` so the first-user fallback is reflected when `defaultUserId` is absent or stale. New test in `config.test.ts` covers it.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc — replaced with `formatAccountLabel` (used by `use` + `remove` with consistent `chalk.dim` styling). Then in abe0fc8 inlined as a template literal at the two call sites since a 1-line helper with 2 callers wasn't earning its keep.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. Extracted `findAccountInStore(store, ref)` into `src/lib/auth-provider.ts` alongside `matchTwistAccount`. `src/commands/auth/store-wrap.ts` now reuses it, so `tw auth logout <wrong>` gets the same `Run: tw account list` hint.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. Account handlers now import `ViewOptions` from `src/lib/options.ts`; the ad-hoc type is gone.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. `src/commands/account/` is now a folder: `index.ts` is wiring only, with `list.ts` / `current.ts` / `use.ts` / `remove.ts` exporting one handler each, plus `helpers.ts` for `assertV2Available`.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. Both `use.ts` and `remove.ts` now resolve the canonical account via `findAccountInStore(store, ref)` and pass `account.id` to `store.setDefault` / `store.clear`, so the store mutates exactly the account that was validated.

src/commands/account/index.ts
scottlovegrove: Fixed in 85097bc. `current.ts` checks for empty `id`/`label` on the snapshot and renders a dedicated "legacy single-user session (pre-multi-account)" notice (or `{source:"legacy"}` in machine output) instead of an account with blank fields. New tests cover both branches.

src/commands/account/account.test.ts
scottlovegrove: Fixed in 85097bc and later condensed via `it.each` in abe0fc8: `emits {source:"env"} in %s mode without touching store.active` runs for both `--json` and `--ndjson` and asserts `storeMocks.active` is never called.

src/commands/account/account.test.ts
scottlovegrove: Fixed in 85097bc. The display-name test now asserts the observable outcome — Alan in the rendered output, no Ellie, single `setDefault` call — rather than the raw ref string. The companion by-id test still asserts `setDefault('1')` because that's the canonical-id contract from comment #7 above.

src/lib/skills/content.ts
scottlovegrove: Fixed in 85097bc, then condensed in abe0fc8. The block now reads `tw account [list|current|use <ref>|remove <ref>]   # all support --json/--ndjson` plus a note documenting `current`'s discriminated payload shape (`{source:"config"} | {source:"env"} | {source:"legacy"}`). Skill is in sync.

scottlovegrove:
@doistbot /review

src/commands/auth/store-wrap.ts
```diff
@@ -20,13 +19,7 @@ export function withUserRefAware(
...
         clear: async (ref?: AccountRef) => {
             if (ref === undefined && requestedRef !== undefined) {
-                const records = await store.list()
-                if (!records.some(({ account }) => matchTwistAccount(account, requestedRef))) {
-                    throw new CliError(
-                        'ACCOUNT_NOT_FOUND',
-                        `No stored account matches "${requestedRef}".`,
-                    )
-                }
+                await findAccountInStore(store, requestedRef)
```
doistbot: [P1] `findAccountInStore` resolves the canonical account, but its return value is discarded here. The unresolved `requestedRef` (which could be a fuzzy display name) is still passed to `store.clear()` on line 24. Since cli-core's token store only matches by exact ID, this will result in a silent no-op when logging out via a fuzzy name. Please capture the resolved account and pass `account.id` to `store.clear()` instead.

src/commands/account/current.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    emitView(
+        options,
+        {
+            id: account.id,
+            label: account.label,
+            authMode: account.authMode,
+            authScope: account.authScope || undefined,
+            source: 'config',
```
doistbot, line 42: [P2] This hard-codes `source: "config"` for every non-empty snapshot, but `createTwistTokenStore.active()` also synthesizes legacy v1 snapshots with populated `id`/`label` when `authUserId` / `authUserName` still exist on disk. In that migration-pending state, `tw account current` will claim the active account is config-backed even though `account list/use/remove` are blocked as legacy-only. Preserve the legacy/v2 distinction here—e.g. thread a source discriminator out of `active()` or check `isLegacyAuthActive()` before rendering—so identified legacy sessions emit `source: "legacy"` too.

src/lib/__fixtures__/accounts.ts
```diff
@@ -0,0 +1,23 @@
...
+// shape (`label`, required `authScope`). ACCOUNT_* fixtures spread STORED_*
+// + add the account-only fields so a future field only needs to land once.
+
+export const STORED_ALAN: StoredUser = { id: '1', name: 'Alan Grant', authMode: 'read-write' }
+export const STORED_ELLIE: StoredUser = { id: '2', name: 'Ellie Sattler', authMode: 'read-only' }
+
+export const ACCOUNT_ALAN: TwistAccount = {
+    ...STORED_ALAN,
+    label: STORED_ALAN.name,
+    authMode: 'read-write',
```
doistbot, line 12: [P3] `ACCOUNT_*` already spreads `STORED_*`, which includes `authMode`. Re-declaring the same literal here creates a second source of truth inside the shared fixtures. Drop the redundant `authMode` override and only add the account-only fields (`label`, `authScope`) so config/account tests stay coupled to one definition.

src/lib/__fixtures__/accounts.ts
```diff
@@ -0,0 +1,23 @@
+import type { TwistAccount } from '../auth-provider.js'
```
doistbot, line 1: [P2] This test fixture file will be compiled into the production `dist/` folder because `tsconfig.build.json` only excludes `src/**/*.test.ts`, `src/**/*.spec.ts`, and `src/__mocks__`. Consider either moving this file into `src/__mocks__/` (or a dedicated test directory if you add it to the exclude list) or updating `tsconfig.build.json` to exclude `src/**/__fixtures__`.

src/lib/__fixtures__/accounts.ts
```diff
@@ -0,0 +1,23 @@
...
+
+// StoredUser is the on-disk shape (`name`); TwistAccount is the in-memory
+// shape (`label`, required `authScope`). ACCOUNT_* fixtures spread STORED_*
+// + add the account-only fields so a future field only needs to land once.
+
+export const STORED_ALAN: StoredUser = { id: '1', name: 'Alan Grant', authMode: 'read-write' }
+export const STORED_ELLIE: StoredUser = { id: '2', name: 'Ellie Sattler', authMode: 'read-only' }
+
+export const ACCOUNT_ALAN: TwistAccount = {
+    ...STORED_ALAN,
```
doistbot: [P2] These `TwistAccount` fixtures are built by spreading `StoredUser`, so they carry a hidden runtime `name` property even though the type says only `label` exists. That weakens the test boundary: code can accidentally read `account.name` and still pass. Please build the `ACCOUNT_*` fixtures from explicit account fields (or `makeTwistAccount(...)`) instead of spreading the stored-user shape.

src/commands/config/config.test.ts
```diff
@@ -273,6 +274,82 @@ describe('config view', () => {
         consoleSpy.mockRestore()
     })
 
+    it('renders an "Authenticated accounts" block from config.users with default marker', async () => {
+        presentConfig({ users: [STORED_ALAN, STORED_ELLIE], defaultUserId: '2' })
+        mockToken('secure-store')
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
```
doistbot, line 280: [P3] The local `consoleSpy` declaration here (and in the tests below) shadows the suite's shared `consoleSpy` and requires manual `.mockRestore()` calls at the end of each block. Relying on the outer spy managed by the existing `beforeEach`/`afterEach` hooks will remove this boilerplate.

src/commands/account/account.test.ts
```diff
@@ -0,0 +1,290 @@
...
+    afterEach(() => {
+        consoleSpy.mockRestore()
+        errorSpy.mockRestore()
+        vi.unstubAllEnvs()
+    })
+
+    const stdout = () => consoleSpy.mock.calls.map((c: unknown[]) => String(c[0])).join('\n')
+
+    describe('list', () => {
+        it('renders all stored accounts with the default marker', async () => {
```
doistbot, line 79: [P2] The new suite only exercises the explicit `tw account list` path. This PR also introduces `list` as the default subcommand (`tw account`), which is a separate Commander wiring contract and currently untested. Please add a bare `parseAsync(['node', 'tw', 'account'])` case so a future routing regression is caught.

src/commands/config/config.test.ts
```diff
@@ -273,6 +274,82 @@ describe('config view', () => {
...
+        expect(output).toContain('id:2')
+        expect(output).toContain('Ellie Sattler')
+        const bobLine = output.split('\n').find((l) => l.includes('Ellie Sattler')) ?? ''
+        const adaLine = output.split('\n').find((l) => l.includes('Alan Grant')) ?? ''
+        expect(bobLine).toContain('*')
+        expect(adaLine).not.toContain('*')
+        consoleSpy.mockRestore()
+    })
+
+    it('marks the first stored account as default when defaultUserId is missing', async () => {
```
doistbot, line 297: [P2] This regression test only covers the "missing `defaultUserId`" branch, but the change in `view.ts` also relies on the stale-id fallback (`defaultUserId` points at no stored user). That stale case was the risky path in the bug fix and it still is not exercised here. Add a case with `defaultUserId` set to a removed/nonexistent id and assert the first stored account gets the marker.

doistbot (commented):
This PR introduces the new `tw account` command tree and updates the config view to cleanly support the multi-account experience. The implementation provides a solid foundation for managing multiple stored accounts while safely masking sensitive tokens. A few adjustments are needed to ensure reliable account removal via fuzzy names, properly distinguish legacy configurations in the active account display, and refine test fixtures and coverage for edge cases.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F233&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0eb4c3f0-51ee-11f1-89e7-645e3609d309%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0eb4c3f0-51ee-11f1-89e7-645e3609d309&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/store-wrap.ts
scottlovegrove: Fixed in 1efde61. `clear` now captures the resolved account from `findAccountInStore` and passes `account.id` to `store.clear`, matching the pattern in `use`/`remove`. (Note: cli-core's keyring `clear` does in fact go through our `matchAccount: matchTwistAccount` so a fuzzy ref would have resolved — but single-sourcing the matching site is still the right shape.)

src/commands/account/current.ts
scottlovegrove: Fixed in 1efde61. `current.ts` now consults `isLegacyAuthActive()` in addition to the empty-id check, so a populated legacy snapshot (v1 `authUserId` / `authUserName` on disk, migration not yet complete) renders as `source:"legacy"` like the symmetric `list/use/remove` block. New test pins the populated-legacy path.

src/lib/__fixtures__/accounts.ts
scottlovegrove: Fixed in 1efde61 — `ACCOUNT_*` are now built explicitly (`id`, `label`, `authMode`, `authScope`) rather than via spread, so there's a single definition per field. Closes the duplicate-`authMode` concern at the same time as the `name`-leak one below.

src/lib/__fixtures__/accounts.ts
scottlovegrove: Fixed in 1efde61 — added `src/**/__fixtures__` to `tsconfig.build.json`'s exclude. Verified with a clean rebuild: `find dist -path '*fixture*'` is empty.

src/lib/__fixtures__/accounts.ts
scottlovegrove: Fixed in 1efde61. `ACCOUNT_ALAN` / `ACCOUNT_ELLIE` are now built from explicit fields (`id: STORED_ALAN.id, label: STORED_ALAN.name, authMode: 'read-write', authScope: 'user:read'`) — no spread, no hidden `name` field, so the `TwistAccount` type boundary stays honest.

src/commands/config/config.test.ts
scottlovegrove: Skipping — there is no shared `consoleSpy` in this file's outer `describe`; every existing test declares a local one with manual restore. Refactoring the whole file's testing pattern is out of scope for this PR, but agreed the file would benefit from a sweep — happy to do that in a follow-up.

src/commands/account/account.test.ts
scottlovegrove: Fixed in 1efde61 — added `runs by default when no subcommand is given (tw account)` in the `list` describe.

src/commands/config/config.test.ts
scottlovegrove: Fixed in 1efde61 — added `falls back to the first stored account when defaultUserId is stale` (sets `defaultUserId: '999'` against `users: [STORED_ALAN, STORED_ELLIE]` and asserts the marker lands on Alan).

henningmu (commented):
Looks good. It's probably also a good candidate for cli-core

scottlovegrove:
> Looks good. It's probably also a good candidate for cli-core

Yeah, that's the plan. Just need to have a plan in place for todoist-cli because that uses `user` instead of `account`, but should be a simple enough switch

PR #243, opened 2026-05-21
feat(auth): adopt cli-core createDcrProvider for OAuth

scottlovegrove:
## Summary

- Replaces the bespoke ~170-LOC DCR `AuthProvider` (hand-rolled registration, PKCE authorize, HTTP Basic token exchange) with cli-core's `createDcrProvider`.
- Bumps `@doist/cli-core` `0.16.1 → 0.20.0` and adds `oauth4webapi@3.8.6` as a direct dependency (cli-core's optional peer dep, required by the DCR flow).
- `validate` now derives `authMode` / `authScope` from the `runOAuthFlow`-folded `handshake.readOnly` instead of values stashed during a hand-rolled `authorize` — the scope set is a pure function of `readOnly` (see `resolveScopes` in `login.ts`).
- Net **−221 LOC** in `auth-provider.ts` + tests.

## Why `client_secret_post` (not the default `client_secret_basic`)

Twist client_ids contain `_` (e.g. `twd_…`). oauth4webapi's `client_secret_basic` form-url-encodes the HTTP Basic credential per RFC 6749 §2.3.1, turning `_` into `%5F`. Twist's token endpoint doesn't url-decode the credential, so it looks up `twd%5F…` and returns `404 "Integration with client_id not found"`. `client_secret_post` sends the credential in the request **body** via `URLSearchParams`, which preserves the underscore, so the lookup succeeds. (The old bespoke code worked because it sent the credential raw; Twist accepts raw.)

## Test plan

- [x] `npm run type-check` — clean
- [x] `npm test` — 645 pass
- [x] `npm run lint` — clean
- [x] `npm run build` — clean
- [x] Live `tw auth login` against twist.com — full DCR registration + PKCE + token exchange + persist, verified against published cli-core `0.20.0`

### DIFF
--- a/src/commands/auth/auth.test.ts
+++ b/src/commands/auth/auth.test.ts
@@ -437,6 +437,10 @@ describe('auth command', () => {
                     return [{ account: SNAPSHOT_ACCOUNT, isDefault: true }]
                 },
                 async setDefault() {},
+                async setBundle() {},
+                async activeBundle() {
+                    return { account: SNAPSHOT_ACCOUNT, bundle: { accessToken: 'snapshot_token' } }
+                },
                 getLastStorageResult: () => undefined,
                 getLastClearResult: () => undefined,
             }
@@ -535,6 +539,10 @@ describe('auth command', () => {
                     return []
                 },
                 async setDefault() {},
+                async setBundle() {},
+                async activeBundle() {
+                    return null
+                },
                 getLastStorageResult: () => undefined,
                 getLastClearResult: () => undefined,
             }
--- a/src/lib/auth-provider.test.ts
+++ b/src/lib/auth-provider.test.ts
@@ -56,16 +56,12 @@ vi.mock('./config.js', async (importOriginal) => {
 
 import { createWrappedTwistClient } from './api.js'
 import {
-    AUTHORIZATION_URL,
     createTwistAuthProvider,
     matchTwistAccount,
     READ_ONLY_SCOPES,
     READ_WRITE_SCOPES,
-    REGISTRATION_URL,
-    TOKEN_URL,
 } from './auth-provider.js'
 
-const REDIRECT_URI = 'http://127.0.0.1:8766/callback'
 const mockCreateClient = vi.mocked(createWrappedTwistClient)
 const TOKEN_ENV_VAR = 'TWIST_API_TOKEN'
 
@@ -89,8 +85,6 @@ const LEGACY_CONFIG = {
     authScope: 'user:read',
 }
 
-const json = (body: unknown, status = 200) => new Response(JSON.stringify(body), { status })
-
 /** Reset the module-level migration memo for each test by re-importing. */
 async function loadCreateTwistTokenStore(): Promise<
     typeof import('./auth-provider.js').createTwistTokenStore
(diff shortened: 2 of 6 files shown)


doistbot (commented):
This update successfully streamlines the OAuth flow by replacing the custom DCR provider with `@doist/cli-core`'s implementation, significantly reducing boilerplate and improving maintainability. A few edge cases require attention before merging, particularly ensuring the token store's `activeBundle()` method mirrors the `active()` logic to prevent incorrect status reporting for environment tokens and pending migrations. Additional refinements include extracting the duplicated scope resolution logic into a shared helper to prevent drift, adding stricter validation for the untyped `handshake.readOnly` flag to protect local access guards, and preserving a dedicated unit test for the critical `client_secret_post` metadata override.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F243&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A1fd9d300-5536-11f1-863f-c78509c953d1%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A1fd9d300-5536-11f1-863f-c78509c953d1&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/auth/auth.test.ts
```diff
@@ -437,6 +437,10 @@ describe('auth command', () => {
                     return [{ account: SNAPSHOT_ACCOUNT, isDefault: true }]
                 },
                 async setDefault() {},
+                async setBundle() {},
+                async activeBundle() {
```
doistbot, line 441: [P1] These new `activeBundle()` stubs are a hint that real cli-core auth commands now read tokens through `store.activeBundle()`. The production `createTwistTokenStore()` still only customizes `active()`, so the inherited `activeBundle()` will bypass Twist's `TWIST_API_TOKEN` short-circuit and legacy snapshot fallback. After this upgrade, `tw auth status` can incorrectly report `NO_TOKEN` for env-token and migration-pending users unless the store adapter mirrors the `active()` resolution logic in `activeBundle()` too.

src/lib/auth-provider.ts
```diff
@@ -76,192 +75,47 @@ export type TwistAccount = AuthAccount & {
...
-            const hs = asHandshake(handshake)
+        errorHints: AUTH_HINTS,
+        async validate({ token, handshake }) {
+            const readOnly = Boolean(handshake.readOnly)
             const client = createWrappedTwistClient(token)
             const user = await client.users.getSessionUser()
-            return toTwistAccount(user, { authMode: hs.authMode, authScope: hs.authScope })
+            return toTwistAccount(user, {
+                authMode: readOnly ? 'read-only' : 'read-write',
+                authScope: (readOnly ? READ_ONLY_SCOPES : READ_WRITE_SCOPES).join(' '),
```
doistbot: [P2] This re-derives the `readOnly` → `authMode`/`authScope` mapping that's already encoded in `attachTwistLoginCommand`'s `resolveScopes`. Keeping two copies means a future scope change can make the stored account metadata drift from the actual grant requested at login. Please extract a shared helper (for scopes or full auth metadata) and reuse it here and in `login.ts`.

src/lib/auth-provider.ts
```diff
@@ -76,192 +75,47 @@ export type TwistAccount = AuthAccount & {
...
-            }
-
-            return { accessToken: data.access_token }
-        },
-
-        async validateToken({ token, handshake }) {
-            const hs = asHandshake(handshake)
+        errorHints: AUTH_HINTS,
+        async validate({ token, handshake }) {
+            const readOnly = Boolean(handshake.readOnly)
```
doistbot: [P2] `handshake` is an untyped cross-layer record here, and `Boolean(handshake.readOnly)` silently turns a missing/malformed value into `read-write`. Because `authMode` is later used by `ensureWriteAllowed()`, a bad handshake would disable the local read-only guard. Please validate that `readOnly` is actually a boolean (or model the handshake with a typed helper) and fail the login instead of defaulting.

src/lib/auth-provider.test.ts
```diff
@@ -101,138 +95,46 @@ async function loadCreateTwistTokenStore(): Promise<
...
-        })
-
-        // Guard: missing verifier means authorize() was never run — never hits fetch.
-        await expect(
-            provider.exchangeCode({ ...base, handshake: { clientId: 'a', clientSecret: 'b' } }),
-        ).rejects.toMatchObject({ code: 'AUTH_FAILED' })
-    })
-
-    it('validateToken fetches getSessionUser with the new token and returns a narrow TwistAccount', async () => {
+    // Registration / authorize / token-exchange mechanics now live in cli-core's
```
doistbot, line 118: [P2] This narrows the tests too far: `validate` is not the only Twist-specific behavior in the adapter. The `client_secret_post` override passed to `createDcrProvider()` is the critical compatibility workaround in this PR, and after deleting the old flow tests there is no local assertion that we still pass that non-default metadata. Please keep a focused unit test around the `createDcrProvider` options so a future refactor can’t silently fall back to cli-core’s default.

src/commands/auth/auth.test.ts
scottlovegrove: Fixed in c6012f7 (the real bug, good catch). `createTwistTokenStore` now overrides `activeBundle()` with the same env-token + legacy-snapshot resolution as `active()` — both share a new `resolveOverrideSnapshot(ref)` helper so they can't drift. Without it, cli-core's `activeBundle`-based reads (e.g. `tw auth status` with `fetchLive`) inherited the plain v2 lookup and would report env-token / migration-pending users as signed out. Added tests covering the env short-circuit and the v2 delegate path.

src/lib/auth-provider.ts
scottlovegrove: Fixed in c6012f7. Extracted `scopesForReadOnly(readOnly)` in auth-provider.ts; both `attachTwistLoginCommand`'s `resolveScopes` (what we request) and the provider's `validate` (what we record) now call it, so the two can't drift.

src/lib/auth-provider.ts
scottlovegrove: Fixed in c6012f7. `validate` now throws `AUTH_FAILED` when `handshake.readOnly` isn't a boolean, instead of `Boolean(undefined)` silently granting read-write (which would relax `ensureWriteAllowed`). runOAuthFlow always folds a boolean, so this only fires on malformed state — fail closed. Added a test.

src/lib/auth-provider.test.ts
scottlovegrove: Fixed in c6012f7. Added a unit test asserting `createTwistAuthProvider` passes `clientMetadata.tokenEndpointAuthMethod: 'client_secret_post'` to `createDcrProvider` (the mock delegates to the real factory so the returned provider still works). A refactor that drops back to cli-core's default Basic now turns the suite red.

src/lib/auth-provider.ts
```diff
@@ -60,6 +59,16 @@ export const READ_ONLY_SCOPES = [
 
 const AUTH_HINTS = ['Try again: tw auth login', 'Or set TWIST_API_TOKEN environment variable']
 
+/**
+ * The scope set a login is granted, as a pure function of `--read-only`. Single
+ * source of truth shared by `attachTwistLoginCommand`'s `resolveScopes` (what we
+ * request at authorize) and the provider's `validate` (what we record on the
+ * account), so the two can't drift.
+ */
+export function scopesForReadOnly(readOnly: boolean): string[] {
```
scottlovegrove: This function name is not good. It suggests it's for getting readonly scopes, yet still accepts a readonly boolean and could return readwrite scopes 🤦🏻‍♂️ A better name would probably be just `getScopes`.

scottlovegrove: Good call — renamed to `getScopes` in 541c1e4.

rfgamaral (commented):
✅

PR #245, opened 2026-05-22
fix(users): filter removed members from `tw users` by default

lmjabreu:
## Summary

`tw users` was listing users who had been **removed** from the workspace. In `Doist` (ws:1585) this meant 554 entries returned for what is actually a 146-member workspace, polluting `tw users [--json]` output and silently breaking `resolveUserRefs` — any command that takes a name/email user ref (`tw channel add`, `tw groups add-user`, etc.) could resolve to a tombstoned user and then 403 from the backend with `User N is not part of workspace`.

Root cause: Twist's v4 `workspace_users/get?id=<wsid>` endpoint returns both active and removed members with a `removed: boolean` flag. The SDK's `WorkspaceUserSchema` already exposes that field — the CLI just wasn't reading it. The curated `--json` formatter also stripped it, so the staleness was invisible.

Reproducer (before the fix):

```
$ tw users 1585 --json --full | jq '[.[] | select(.removed)] | length'
408
$ tw users 1585 --json | jq '. | map(select(.id == 2342 or .id == 2343))'
[
  { "id": 2342, "name": "Test User1",  "userType": "GUEST" },
  { "id": 2343, "name": "Test User 2", "userType": "USER"  }
]
$ # Both above are removed members; adding them to a channel returns 403
$ # "User 2342 is not part of workspace 1585"
```

After the fix:

```
$ tw users 1585 --json | jq '. | length'
146
$ tw users 1585 --json --include-removed | jq '. | length'
554
$ tw users 1585 --include-removed --search "Test User1"
id:2342  Test User1 <test1@doist.io> [GUEST] [removed]
```

## Changes

| Layer | File | Notes |
|---|---|---|
| API helper | `src/lib/api.ts` | `getWorkspaceUsers(workspaceId, { includeRemoved? })` filters `removed === true` by default. One filter, fixes `tw users`, `resolveUserRefs` ([src/lib/refs.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/lib/refs.ts)), and `resolveNotifyIds` ([src/commands/thread/helpers.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/commands/thread/helpers.ts)) in one shot. |
| Command | `src/commands/user.ts` | New `--include-removed` flag. Human output appends a red `[removed]` chip. |
| JSON | `src/lib/output.ts` | Added `removed` to `USER_ESSENTIAL_FIELDS` so callers can distinguish the two states without `--full`. |
| Tests | `src/commands/user.test.ts` | Two new tests: default omits removed users, flag passes through and chip renders. |

`getUserById`-based callers (e.g. `tw groups view`) were already correct because they only fetch ids that the group's `userIds` list references — removed users aren't there.

## Why a flag, not always filter

Audit/recovery cases occasionally do want the historical roster (e.g. "who used to be in here?"). `--include-removed` covers that without making the common case noisy.

## Side discoveries

None. Scope was kept tight to the diagnosed bug.

## Checks

- [x] `npm run type-check`
- [x] `npm run lint:check`
- [x] `npm test` — 653/653 (2 new)
- [x] Manual end-to-end against workspace 1585 (Doist) — counts above

Follow-up to side-discovery #3 in #244.

### DIFF
--- a/src/commands/user.test.ts
+++ b/src/commands/user.test.ts
@@ -88,17 +88,45 @@ describe('user --json', () => {
         expect(jsonOutput).not.toHaveProperty('shortName')
     })
 
-    it('omits removed users by default and passes includeRemoved: undefined', async () => {
-        apiMocks.getWorkspaceUsers.mockResolvedValueOnce([
-            {
-                id: 1,
-                name: 'Active',
-                email: 'a@x',
-                userType: 'USER',
-                bot: false,
-                removed: false,
-            },
-        ])
+    it('outputs full user fields with --full', async () => {
+        const program = createProgram()
+        const consoleSpy = captureConsole()
+
+        await program.parseAsync(['node', 'tw', 'user', '--json', '--full'])
+
+        expect(consoleSpy).toHaveBeenCalledTimes(1)
+        const jsonOutput = JSON.parse(consoleSpy.mock.calls[0][0])
+        expect(jsonOutput).toHaveProperty('lang', 'en')
+        expect(jsonOutput).toHaveProperty('shortName', 'Jane')
+        expect(jsonOutput).toHaveProperty('defaultWorkspace', 1)
+    })
+})
+
+describe('tw users --include-removed', () => {
+    const active = {
+        id: 1,
+        name: 'Active',
+        email: 'a@x',
+        userType: 'USER',
+        bot: false,
+        removed: false,
+    }
+    const removed = {
+        id: 2,
+        name: 'Ghost',
+        email: 'ghost@x',
+        userType: 'GUEST',
+        bot: false,
+        removed: true,
+    }
+
+    beforeEach(() => {
+        vi.clearAllMocks()
+        apiMocks.getCurrentWorkspaceId.mockResolvedValue(1)
+    })
+
+    it('passes includeRemoved: undefined by default so the SDK applies its default filter', async () => {
+        apiMocks.getWorkspaceUsers.mockResolvedValueOnce([active])
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
 
@@ -111,37 +139,34 @@ describe('user --json', () => {
     })
 
     it('passes includeRemoved: true and annotates removed users in text output', async () => {
-        apiMocks.getWorkspaceUsers.mockResolvedValueOnce([
-            {
-                id: 2,
-                name: 'Ghost',
-                email: 'ghost@x',
-                userType: 'GUEST',
-                bot: false,
-                removed: true,
-            },
-        ])
+        apiMocks.getWorkspaceUsers.mockResolvedValueOnce([active, removed])
         const program = createProgram()
         const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
 
         await program.parseAsync(['node', 'tw', 'users', '--include-removed'])
 
         expect(apiMocks.getWorkspaceUsers).toHaveBeenCalledWith(1, { includeRemoved: true })
-        expect(consoleSpy.mock.calls.flat().join('\n')).toMatch(/\[removed\]/)
+        const lines = consoleSpy.mock.calls.flat().join('\n')
+        expect(lines).toMatch(/id:2.*Ghost.*\[removed\]/)
+        expect(lines).not.toMatch(/id:1.*Active.*\[removed\]/)
 
         consoleSpy.mockRestore()
     })
 
-    it('outputs full user fields with --full', async () => {
+    it('surfaces removed in curated --json output without --full', async () => {
+        apiMocks.getWorkspaceUsers.mockResolvedValueOnce([active, removed])
         const program = createProgram()
-        const consoleSpy = captureConsole()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
 
-        await program.parseAsync(['node', 'tw', 'user', '--json', '--full'])
+        await program.parseAsync(['node', 'tw', 'users', '--include-removed', '--json'])
 
-        expect(consoleSpy).toHaveBeenCalledTimes(1)
-        const jsonOutput = JSON.parse(consoleSpy.mock.calls[0][0])
-        expect(jsonOutput).toHaveProperty('lang', 'en')
-        expect(jsonOutput).toHaveProperty('shortName', 'Jane')
-        expect(jsonOutput).toHaveProperty('defaultWorkspace', 1)
+        const parsed = JSON.parse(consoleSpy.mock.calls[0][0])
+        expect(parsed).toHaveLength(2)
+        expect(parsed[0]).toMatchObject({ id: 1, removed: false })
+        expect(parsed[1]).toMatchObject({ id: 2, removed: true })
+        // Curated, not --full: shortName must not leak in.
+        expect(parsed[0]).not.toHaveProperty('shortName')
+
+        consoleSpy.mockRestore()
     })
 })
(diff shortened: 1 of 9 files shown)


src/commands/user.ts
```diff
@@ -98,6 +99,7 @@ Examples:
         .description('List users in a workspace')
         .option('--workspace <ref>', 'Workspace ID or name')
         .option('--search <text>', 'Filter by name/email')
+        .option('--include-removed', 'Include users who have been removed from the workspace')
```
doistbot, line 102: [P2] Per the repository guidance in `AGENTS.md`, whenever a command, flag, or option is added, `SKILL_CONTENT` in `src/lib/skills/content.ts` must be updated to match. Please add `--include-removed` to the `tw users` section in that file so AI agents know the flag exists.

src/commands/user.test.ts
```diff
@@ -92,6 +92,50 @@ describe('user --json', () => {
         consoleSpy.mockRestore()
     })
 
+    it('omits removed users by default and passes includeRemoved: undefined', async () => {
```
doistbot: [P3] These tests verify standard text output (omitting the `--json` flag and checking for the `[removed]` chip), but they are placed inside the `describe('user --json', ...)` block. Move them out of this block to the section dedicated to standard text output.

doistbot (commented):
This PR thoughtfully addresses a significant issue with reference resolution by ensuring `tw users` filters out removed members by default, while introducing an `--include-removed` flag to access tombstoned users when needed. These changes greatly improve the reliability of workspace user management and correctly utilize the v4 API schema. A few minor adjustments are needed to document the new flag in the AI agent skill content, realign a misplaced test block, and strengthen the test suite to properly verify both the default filtering behavior and the updated JSON payload.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F245&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae08627d0-55c6-11f1-941d-71d784f040d4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae08627d0-55c6-11f1-941d-71d784f040d4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/user.test.ts
doistbot: [P2] This doesn't actually verify the new default-filtering behavior. The mock only returns an active user, so the test still passes if `src/lib/api.ts` stops filtering out `removed` members. Add coverage at the API-helper level, or an integration-style `tw users` test that doesn't stub `getWorkspaceUsers`, with a mixed active/removed payload and an assertion that the removed entry is absent by default.

src/commands/user.test.ts
```diff
@@ -92,6 +92,50 @@ describe('user --json', () => {
...
+
+        await program.parseAsync(['node', 'tw', 'users'])
+
+        expect(apiMocks.getWorkspaceUsers).toHaveBeenCalledWith(1, { includeRemoved: undefined })
+        expect(consoleSpy.mock.calls.flat().join('\n')).not.toMatch(/\[removed\]/)
+
+        consoleSpy.mockRestore()
+    })
+
+    it('passes includeRemoved: true and annotates removed users in text output', async () => {
```
doistbot, line 141: [P2] The machine-readable case introduced by this change is still untested. `src/lib/output.ts` now includes `removed` in essential user fields, but these tests only check text output. Add a `tw users --json --include-removed` assertion that the output contains `removed: true` without `--full`, so a regression in `USER_ESSENTIAL_FIELDS` gets caught.

scottlovegrove (commented):
I'm wondering whether this should be the default of the _SDK_ rather than this. The MCP probably has the same issue :/ 

I think I would prefer this to be the default of the SDK

scottlovegrove:
> I'm wondering whether this should be the default of the _SDK_ rather than this. The MCP probably has the same issue :/
> 
> I think I would prefer this to be the default of the SDK

This has been done and is available in 2.8.1 of the SDK. @lmjabreu ask your agent to pull in that newer SDK, and then we can tidy up this PR. The PR should still make the flag available, but it doesn't need to do any of the filtering.

src/commands/user.ts
lmjabreu: Done — added `tw users --include-removed` to `SKILL_CONTENT` in [src/lib/skills/content.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/lib/skills/content.ts) and regenerated [skills/twist-cli/SKILL.md](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/skills/twist-cli/SKILL.md) via `npm run sync:skill`. `check:skill-sync` is green.

src/commands/user.test.ts
lmjabreu: Moved out — the two new tests now live in their own `describe('tw users --include-removed', ...)` block at the bottom of [src/commands/user.test.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/commands/user.test.ts), separate from the `user --json` describe.

lmjabreu: Good catch. The filter has now moved into the SDK (2.8.1) and `getWorkspaceUsers` is a pure pass-through, so the meaningful contract to pin is "the wrapper forwards `includeRemoved` to the SDK unchanged." Added [src/lib/api.test.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/lib/api.test.ts) which mocks `TwistApi` and asserts both calls — `getWorkspaceUsers(1585)` invokes the SDK with `includeRemoved: undefined`, and `getWorkspaceUsers(1585, { includeRemoved: true })` forwards `true`. If someone drops the forwarding, this fails.

src/commands/user.test.ts
lmjabreu: Added — the new `surfaces removed in curated --json output without --full` test in [src/commands/user.test.ts](https://github.com/Doist/twist-cli/blob/lmjabreu/confident-lalande-769bec/src/commands/user.test.ts) feeds a mixed active/removed payload, runs `tw users --include-removed --json` (no `--full`), and asserts `removed: true` is present on the removed row, `removed: false` on the active row, and `shortName` (a non-essential field) is absent. A regression in `USER_ESSENTIAL_FIELDS` would now fail this test.

lmjabreu:
@scottlovegrove done — bumped to `@doist/twist-sdk@2.8.1` and reduced `getWorkspaceUsers` to a pass-through that forwards `includeRemoved`. Verified against ws:1585: default returns 146 active users (was 554 pre-fix), `--include-removed` returns 554 with the `removed` field intact in curated JSON. The four doistbot review threads are addressed inline + resolved.

scottlovegrove (approved):

PR #246, opened 2026-05-22
feat(channel): add create, delete, archive, unarchive subcommands

lmjabreu:
## Summary

- Adds `tw channel create`, `tw channel delete`, `tw channel archive`, `tw channel unarchive` — wraps the four SDK endpoints that were already exposed but not surfaced in the CLI. Discovered as a gap while landing channel-membership commands in [#244](https://github.com/Doist/twist-cli/pull/244); end-to-end channel testing previously required the web UI.
- `create` supports `--workspace`, `--description`, `--private` (default public), `--dry-run`, `--json`, `--full`.
- `delete` requires `--yes` to mutate, with structured `MISSING_YES_FLAG` CliError in `--json` mode, and a clear `FORBIDDEN` CliError when the Twist API returns 403 (channel deletion is typically restricted to workspace admins).
- `archive` / `unarchive` mirror the same shape with `--dry-run` and `--json`.
- Skill quick-reference + `SKILL.md` regenerated.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm test` — 667/667 passing (17 new tests across the four commands)
- [x] Live smoke against the Doist workspace: create → verify in list → archive → list `--state archived` → unarchive → re-archive
- [x] Live-verified the `FORBIDDEN` error message on `delete --yes` (Twist returns 403 for non-admins, as expected)
- [ ] Live-verified successful `delete` — blocked by workspace admin policy; unit-tested only

### DIFF
--- a/src/commands/channel/archive.ts
+++ b/src/commands/channel/archive.ts
@@ -1,0 +1,56 @@
+import { archiveChannel, getCurrentWorkspaceId, unarchiveChannel } from '../../lib/api.js'
+import type { MutationOptions } from '../../lib/options.js'
+import { formatJson, printDryRun } from '../../lib/output.js'
+import { resolveChannelRef } from '../../lib/refs.js'
+
+type ArchiveChannelOptions = MutationOptions
+
+export async function archiveChannelCommand(
+    ref: string,
+    options: ArchiveChannelOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const channel = await resolveChannelRef(ref, workspaceId)
+
+    if (options.dryRun) {
+        printDryRun('archive channel', {
+            Channel: `${channel.name} (id:${channel.id})`,
+            'Currently archived': channel.archived ? 'yes' : 'no',
+        })
+        return
+    }
+
+    await archiveChannel(channel.id)
+
+    if (options.json) {
+        console.log(formatJson({ id: channel.id, archived: true }))
+        return
+    }
+
+    console.log(`Channel "${channel.name}" (id:${channel.id}) archived.`)
+}
+
+export async function unarchiveChannelCommand(
+    ref: string,
+    options: ArchiveChannelOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const channel = await resolveChannelRef(ref, workspaceId)
+
+    if (options.dryRun) {
+        printDryRun('unarchive channel', {
+            Channel: `${channel.name} (id:${channel.id})`,
+            'Currently archived': channel.archived ? 'yes' : 'no',
+        })
+        return
+    }
+
+    await unarchiveChannel(channel.id)
+
+    if (options.json) {
+        console.log(formatJson({ id: channel.id, archived: false }))
+        return
+    }
+
+    console.log(`Channel "${channel.name}" (id:${channel.id}) unarchived.`)
+}
--- a/src/commands/channel/archive.ts
+++ b/src/commands/channel/archive.ts
@@ -1,56 +1,57 @@
 import { archiveChannel, getCurrentWorkspaceId, unarchiveChannel } from '../../lib/api.js'
 import type { MutationOptions } from '../../lib/options.js'
 import { formatJson, printDryRun } from '../../lib/output.js'
-import { resolveChannelRef } from '../../lib/refs.js'
+import { resolveChannelRef, resolveWorkspaceRef } from '../../lib/refs.js'
 
-type ArchiveChannelOptions = MutationOptions
+type ArchiveChannelOptions = MutationOptions & { workspace?: string }
 
-export async function archiveChannelCommand(
+async function setArchiveState(
     ref: string,
     options: ArchiveChannelOptions,
+    archive: boolean,
 ): Promise<void> {
-    const workspaceId = await getCurrentWorkspaceId()
+    const action = archive ? 'archive' : 'unarchive'
+    const workspaceId = options.workspace
+        ? (await resolveWorkspaceRef(options.workspace)).id
+        : await getCurrentWorkspaceId()
     const channel = await resolveChannelRef(ref, workspaceId)
 
     if (options.dryRun) {
-        printDryRun('archive channel', {
+        printDryRun(`${action} channel`, {
             Channel: `${channel.name} (id:${channel.id})`,
             'Currently archived': channel.archived ? 'yes' : 'no',
         })
         return
     }
 
-    await archiveChannel(channel.id)
+    if (channel.archived !== archive) {
+        if (archive) {
+            await archiveChannel(channel.id)
+        } else {
+            await unarchiveChannel(channel.id)
+        }
+    }
 
     if (options.json) {
-        console.log(formatJson({ id: channel.id, archived: true }))
+        console.log(formatJson({ id: channel.id, archived: archive }))
         return
     }
 
-    console.log(`Channel "${channel.name}" (id:${channel.id}) archived.`)
+    const verb = archive ? 'archived' : 'unarchived'
+    const noop = channel.archived === archive ? ' (already in target state)' : ''
+    console.log(`Channel "${channel.name}" (id:${channel.id}) ${verb}${noop}.`)
+}
+
+export async function archiveChannelCommand(
+    ref: string,
+    options: ArchiveChannelOptions,
+): Promise<void> {
+    await setArchiveState(ref, options, true)
 }
 
 export async function unarchiveChannelCommand(
     ref: string,
     options: ArchiveChannelOptions,
 ): Promise<void> {
-    const workspaceId = await getCurrentWorkspaceId()
-    const channel = await resolveChannelRef(ref, workspaceId)
-
-    if (options.dryRun) {
-        printDryRun('unarchive channel', {
-            Channel: `${channel.name} (id:${channel.id})`,
-            'Currently archived': channel.archived ? 'yes' : 'no',
-        })
-        return
-    }
-
-    await unarchiveChannel(channel.id)
-
-    if (options.json) {
-        console.log(formatJson({ id: channel.id, archived: false }))
-        return
-    }
-
-    console.log(`Channel "${channel.name}" (id:${channel.id}) unarchived.`)
+    await setArchiveState(ref, options, false)
 }
--- a/src/commands/channel/channel.test.ts
+++ b/src/commands/channel/channel.test.ts
@@ -412,7 +412,7 @@ describe('tw channel create', () => {
     it('resolves --workspace ref when provided', async () => {
         refsMocks.resolveWorkspaceRef.mockResolvedValue({ id: 42, name: 'Other' })
         const program = createProgram()
-        vi.spyOn(console, 'log').mockImplementation(() => {})
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
 
         await program.parseAsync([
             'node',
@@ -431,6 +431,8 @@ describe('tw channel create', () => {
             description: undefined,
             public: true,
         })
+
+        consoleSpy.mockRestore()
     })
 
     it('passes --description and --private through to createChannel', async () => {
@@ -438,7 +440,7 @@ describe('tw channel create', () => {
             createChannel(999, 'Leadership', { public: false, description: 'Internal' }),
         )
         const program = createProgram()
-        vi.spyOn(console, 'log').mockImplementation(() => {})
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
 
         await program.parseAsync([
             'node',
@@ -457,6 +459,8 @@ describe('tw channel create', () => {
             description: 'Internal',
             public: false,
         })
+
+        consoleSpy.mockRestore()
     })
 
     it('does not call the API on --dry-run', async () => {
@@ -491,6 +495,30 @@ describe('tw channel create', () => {
         const output = JSON.parse(consoleSpy.mock.calls[0][0])
         expect(output.id).toBe(123)
         expect(output.name).toBe('Engineering')
+
+        consoleSpy.mockRestore()
+    })
+
+    it('forwards --full so JSON output includes all channel fields', async () => {
+        apiMocks.createChannel.mockResolvedValue(
+            createChannel(123, 'Engineering', { description: 'Eng channel', creator: 9 }),
+        )
+        const program = createProgram()
+        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+
+        await program.parseAsync([
+            'node',
+            'tw',
+            'channel',
+            'create',
+            'Engineering',
+            '--json',
+            '--full',
+        ])
+
+        const output = JSON.parse(consoleSpy.mock.calls[0][0])
+        // The default-shape filter drops `public`/`creator`; --full keeps them.
+        expect(output).toMatchObject({ id: 123, name: 'Engineering', public: true, creator: 9 })
 
         consoleSpy.mockRestore()
     })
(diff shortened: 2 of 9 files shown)


src/commands/channel/delete.ts
```diff
@@ -0,0 +1,67 @@
...
+        console.log('Use --yes to confirm.')
+        return
+    }
+
+    try {
+        await deleteChannel(channel.id)
+    } catch (error) {
+        if (isForbidden(error)) {
+            throw new CliError(
+                'FORBIDDEN',
```
doistbot, line 45: [P2] The `FORBIDDEN` error code is used here but hasn't been added to the `ErrorCode` union in `src/lib/errors.ts`. As per `AGENTS.md`, new error codes must be added to that type under the appropriate category.

src/commands/channel/index.ts
```diff
@@ -86,4 +89,74 @@ Notes:
...
+Examples:
+  tw channel create "Engineering"
+  tw channel create "Leadership" --private
+  tw channel create "Marketing" --description "Marketing team channel"
+  tw channel create "Design" --private --json`,
+        )
+        .action(createChannelCommand)
+
+    channel
+        .command('delete <channel-ref>')
```
doistbot, line 118: [P2] Unlike `channel create` and `channel threads`, these new mutation subcommands don't offer any workspace override. Their handlers resolve refs against `getCurrentWorkspaceId()`, so a valid channel URL from another workspace currently fails with `resolveChannelRef()`'s "Pass the matching workspace-ref" hint even though this command family has no way to pass one. Add `--workspace <ref>` here (and on archive/unarchive) or avoid surfacing that hint for commands that can't honor it.

src/commands/channel/delete.ts
```diff
@@ -0,0 +1,67 @@
...
+        (error as { httpStatusCode: number }).httpStatusCode === 403
+    )
+}
+
+export async function deleteChannelCommand(
+    ref: string,
+    options: DeleteChannelOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
+    const channel = await resolveChannelRef(ref, workspaceId)
```
doistbot, line 24: [P2] `resolveChannelRef()` only does name matching against `getChannels({ workspaceId })`, which this codebase treats as the joined-channel set. That means `tw channel delete "Foo"` can't resolve a public channel the admin hasn't joined, and archived channels can't be deleted by name at all. The same lookup pattern is used in archive/unarchive. For these management commands, use a broader lookup (public + archived) or tighten the accepted-ref docs/examples.

src/lib/skills/content.ts
```diff
@@ -221,6 +221,14 @@ tw channel threads <ref> --since 2026-01-01 # Filter by last-updated date (ISO)
 tw channel threads <ref> --limit 20         # Max threads per page (default: 50)
 tw channel threads <ref> --limit 20 --cursor <cursor-from-prev> # Paginate
 tw channel threads <ref> --json  # { results, nextCursor } with isUnread + url
+tw channel create "Name"         # Create a new (public) channel in the current workspace
```
doistbot: [P3] This quick-reference line is now inaccurate: `tw channel create` isn't limited to the current workspace because the command also supports `--workspace <ref>`. Please reflect that in the skill content/examples so the agent-facing command reference stays in sync.

src/commands/channel/archive.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (options.json) {
+        console.log(formatJson({ id: channel.id, archived: true }))
+        return
+    }
+
+    console.log(`Channel "${channel.name}" (id:${channel.id}) archived.`)
+}
+
+export async function unarchiveChannelCommand(
```
doistbot: [P2] `unarchiveChannelCommand` heavily duplicates the resolution, dry-run, execution, and JSON formatting logic of `archiveChannelCommand`. Consider extracting a shared helper (e.g., `setArchiveState(ref, options, archive: boolean)`) to handle both flows.

src/commands/channel/delete.ts
```diff
@@ -0,0 +1,67 @@
+import { deleteChannel, getCurrentWorkspaceId } from '../../lib/api.js'
+import { CliError } from '../../lib/errors.js'
+import type { MutationOptions } from '../../lib/options.js'
+import { formatJson, printDryRun } from '../../lib/output.js'
+import { resolveChannelRef } from '../../lib/refs.js'
+
+type DeleteChannelOptions = MutationOptions & { yes?: boolean }
+
+function isForbidden(error: unknown): boolean {
```
doistbot: [P2] The `isForbidden` check reinvents structural validation for API errors. You can simplify this by importing `TwistRequestError` from `@doist/twist-sdk` and checking `error instanceof TwistRequestError && error.httpStatusCode === 403` (matching the pattern in `src/commands/auth/status.ts`). Note: this will require updating the corresponding test mock to throw `new TwistRequestError(...)` instead of a plain error with an assigned property.

src/commands/channel/delete.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    if (options.dryRun) {
+        printDryRun('delete channel', {
+            Channel: `${channel.name} (id:${channel.id})`,
+            Visibility: channel.public ? 'public' : 'private',
+        })
+        return
+    }
+
+    if (!options.yes) {
```
doistbot, line 34: [P2] This `--yes` gate is now another copy of the confirmed-delete flow already used in `groups/delete.ts` and `thread/delete.ts` (`MISSING_YES_FLAG` in JSON mode, preview text otherwise). Extracting that behavior into a shared helper would avoid a third near-identical branch and keep delete UX aligned across commands.

src/commands/channel/archive.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    if (options.dryRun) {
+        printDryRun('archive channel', {
+            Channel: `${channel.name} (id:${channel.id})`,
+            'Currently archived': channel.archived ? 'yes' : 'no',
+        })
+        return
+    }
+
+    await archiveChannel(channel.id)
```
doistbot: [P2] Since `resolveChannelRef` already fetches the channel and its state, skip the API call if the channel is already archived:

```ts
if (!channel.archived) {
    await archiveChannel(channel.id)
}
```
This avoids a redundant network request and spinner flicker.

src/commands/channel/delete.ts
```diff
@@ -0,0 +1,67 @@
...
+        'httpStatusCode' in error &&
+        (error as { httpStatusCode: number }).httpStatusCode === 403
+    )
+}
+
+export async function deleteChannelCommand(
+    ref: string,
+    options: DeleteChannelOptions,
+): Promise<void> {
+    const workspaceId = await getCurrentWorkspaceId()
```
doistbot: [P2] This does `getCurrentWorkspaceId()` and `resolveChannelRef()` before the `--json`/`--yes` guard below. In the `tw channel delete ... --json` without `--yes` case, the command always throws `MISSING_YES_FLAG`, so these lookups are wasted and can still hit config/API reads on an error path. Move the `!options.yes && options.json` check before resolving the workspace/channel, and only do the lookup when you need a preview or are actually deleting.

src/commands/channel/channel.test.ts
```diff
@@ -401,3 +409,303 @@ describe('channels list', () => {
...
+        expect(consoleSpy.mock.calls[0][0]).toContain('Engineering')
+        expect(consoleSpy.mock.calls[0][0]).toContain('public')
+
+        consoleSpy.mockRestore()
+    })
+
+    it('resolves --workspace ref when provided', async () => {
+        refsMocks.resolveWorkspaceRef.mockResolvedValue({ id: 42, name: 'Other' })
+        const program = createProgram()
+        vi.spyOn(console, 'log').mockImplementation(() => {})
```
doistbot: [P2] `console.log` is mocked here but not restored, which leaves it silenced for subsequent tests (since Vitest does not auto-restore mocks in this project's setup). Capture the spy as `const consoleSpy = ...` and call `consoleSpy.mockRestore()` at the end of the test, matching the pattern in adjacent tests.

src/commands/channel/channel.test.ts
```diff
@@ -401,3 +409,303 @@ describe('channels list', () => {
...
+
+        expect(apiMocks.createChannel).not.toHaveBeenCalled()
+        const text = consoleSpy.mock.calls.map((c) => String(c[0])).join('\n')
+        expect(text).toContain('create channel')
+        expect(text).toContain('private')
+
+        consoleSpy.mockRestore()
+    })
+
+    it('outputs created channel as JSON', async () => {
```
doistbot, line 488: [P2] This is the only machine-output test for `channel create`, but it never exercises the new `--json --full` path. Because channel JSON is field-filtered by default, a regression that stops forwarding `options.full` would still pass here. Add a `--json --full` assertion for a non-essential field like `public` or `archived`.

doistbot (commented):
This PR successfully bridges a crucial functionality gap by introducing subcommands for creating, deleting, archiving, and unarchiving channels. These additions will significantly improve the CLI's utility for workspace administration and end-to-end testing. There are a few opportunities for refinement, specifically around extracting shared logic for delete confirmations and archive toggling, broadening the channel resolution bounds to include unjoined or archived channels, and addressing minor gaps in test coverage, error handling, and agent documentation.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F246&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9cd9f630-55ce-11f1-9355-787aec697626%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9cd9f630-55ce-11f1-9355-787aec697626&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/channel/channel.test.ts
```diff
@@ -401,3 +409,303 @@ describe('channels list', () => {
...
+        await program.parseAsync(['node', 'tw', 'channel', 'delete', 'Engineering', '--yes'])
+
+        expect(refsMocks.resolveChannelRef).toHaveBeenCalledWith('Engineering', 1)
+        expect(apiMocks.deleteChannel).toHaveBeenCalledWith(500)
+        expect(consoleSpy.mock.calls[0][0]).toContain('Engineering')
+
+        consoleSpy.mockRestore()
+    })
+
+    it('does not delete on --dry-run', async () => {
```
doistbot, line 566: [P2] This only checks `--dry-run` without `--yes`, which is already non-mutating because the command exits on missing confirmation. The safety-sensitive case is `--yes --dry-run`: if the guard order changes, this test still passes while a preview starts deleting channels. Please add that combination and assert `deleteChannel` is not called.

src/lib/api.ts
```diff
@@ -341,6 +344,31 @@ export async function removeUsersFromGroup(id: number, userIds: number[]): Promi
     await client.groups.removeUsers({ id, userIds })
 }
 
+export async function createChannel(args: {
+    workspaceId: number
+    name: string
+    description?: string
+    public?: boolean
+}): Promise<Channel> {
```
scottlovegrove: Why aren't we making use of the SDK type for `createChannel` since we're just passing the args straight into the client anyway.

src/commands/channel/delete.ts
lmjabreu: Added `FORBIDDEN` to the `ErrorCode` union in `src/lib/errors.ts` in 95817ad.

src/commands/channel/index.ts
lmjabreu: Added `--workspace <ref>` to `delete`, `archive`, and `unarchive` in 95817ad — matches the shape of `create` and `channel threads`.

src/commands/channel/delete.ts
lmjabreu: Agreed, but this is the existing behaviour of `resolveChannelRef` from before this PR — it bites every command that uses it, not just these. I've drafted a focused follow-up that widens the name-resolution to also cover public and archived channels (and removes the temporary `pass id: for archived channels` signpost from the `unarchive` help). Keeping this PR scoped to the create/delete/archive surface so the fix can land as one self-contained change. Leaving this thread open until that PR lands.

src/lib/skills/content.ts
lmjabreu: Fixed in 95817ad — quick-reference now mentions `--workspace` and uses "default/current workspace" wording. `SKILL.md` regenerated.

src/commands/channel/archive.ts
lmjabreu: Extracted `setArchiveState(ref, options, archive)` in `archive.ts` (95817ad). Both subcommands are now one-line forwarders.

src/commands/channel/delete.ts
lmjabreu: Switched to `error instanceof TwistRequestError && error.httpStatusCode === 403` in 95817ad, and reworked the test to throw a real `TwistRequestError` instead of an assigned-property plain Error.

src/commands/channel/delete.ts
lmjabreu: Agreed — this is the right time to extract, but it's a cross-cutting refactor of `groups/delete.ts`, `thread/delete.ts`, and now `channel/delete.ts`. Out of scope for this PR. Happy to do it next as its own focused change — let me know if you'd rather block this PR on it.

src/commands/channel/archive.ts
lmjabreu: Done in 95817ad via `setArchiveState` — skips the API call when `channel.archived === archive` and surfaces `(already in target state)` in the human output. Added a no-op assertion test.

src/commands/channel/delete.ts
lmjabreu: Reordered in 95817ad — `!options.yes && options.json && !options.dryRun` now short-circuits to `MISSING_YES_FLAG` before any workspace/channel lookup. Added a test asserting `resolveChannelRef` and `getCurrentWorkspaceId` are not called on that path.

src/commands/channel/channel.test.ts
lmjabreu: Fixed in 95817ad — restored all unrestored `console.log` spies in the `tw channel create` describe block.

src/commands/channel/channel.test.ts
lmjabreu: Added in 95817ad — new test "forwards --full so JSON output includes all channel fields" asserts that default-filtered fields (`public`, `creator`) are present when `--full` is passed.

src/commands/channel/channel.test.ts
lmjabreu: Added in 95817ad — new test "does not delete when --yes is combined with --dry-run" asserts `deleteChannel` is not called on that path.

src/lib/api.ts
lmjabreu: Good catch — switched to `CreateChannelArgs` from `@doist/twist-sdk` in 95817ad. Wrapper is now just `(args: CreateChannelArgs) => client.channels.createChannel(args)`.

scottlovegrove (approved):

PR #247, opened 2026-05-22
refactor(errors): centralize 403 FORBIDDEN translation in wrapResult

lmjabreu:
## Summary

- Lift the local `isForbidden` helper from `src/commands/channel/delete.ts` into the central `wrapResult` handler in `src/lib/api.ts`. Every SDK call now gets uniform 403 → `CliError('FORBIDDEN')` translation, not just channel deletion. Future commands (kick user, group delete, etc.) inherit the friendly message for free.
- New `isForbidden(error)` predicate in `src/lib/errors.ts` sits next to `isInsufficientScope`. Both delegate to a shared `hasTwistStatusCode(error, status)` type predicate so the shape narrowing lives in one place, with no `as` casts. JSDoc documents the precedence rule: callers must test `isInsufficientScope` first so OAuth-scope 403s keep their dedicated `INSUFFICIENT_SCOPE` code and hints; `isForbidden` is the catch-all.
- FORBIDDEN end-to-end translation test in `src/lib/api.test.ts` exercises the real proxy + `wrapResult` flow by mocking `@doist/twist-sdk`. Covers FORBIDDEN translation, INSUFFICIENT_SCOPE precedence, and non-403 pass-through (asserted by object identity, not just message).

**Hint copy is call-agnostic.** The central message reads `Twist refused this action: 403 Forbidden.` with hints `You may not have permission for this action` and `Contact your workspace admin, or re-authenticate with \`tw auth login\` if your token looks wrong`. #246's local message included the channel name; uniform coverage is the deliberate tradeoff. If per-command flavour matters later, we can plumb the spinner config's command name into the central message.

**Follow-ups (not this PR):** the same central-translation pattern should eventually cover 404 (`NOT_FOUND`-style with resource hint), 429 (rate-limit with retry-after), and 5xx (transient retry with backoff). Each warrants its own PR with its own behaviour decision. Batch responses (`assertBatchData` / `getOptionalBatchData`) also bypass `wrapResult` and would need a separate 403 check; tracked for the same follow-up.

## Test plan

- [x] `npm run type-check`
- [x] `npm run lint`
- [x] `npm test` (710 passed)
- [x] Manual smoke: `tw channel delete id:837190 --yes` against the parked admin-only channel surfaces the new friendly FORBIDDEN message (verified after `tw channel unarchive`, then re-archived).

### DIFF
--- a/src/lib/api.test.ts
+++ b/src/lib/api.test.ts
@@ -1,3 +1,4 @@
+import { TwistRequestError } from '@doist/twist-sdk'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
 // Hoisted mocks — shared across both describe blocks.
@@ -82,10 +83,7 @@ describe('wrapResult — central 403 translation', () => {
 
     it('translates a plain 403 into a FORBIDDEN CliError', async () => {
         sdkMocks.deleteChannel.mockRejectedValueOnce(
-            Object.assign(new Error('Request failed with status 403'), {
-                httpStatusCode: 403,
-                responseData: {},
-            }),
+            new TwistRequestError('Request failed with status 403', 403, {}),
         )
 
         const { createWrappedTwistClient } = await import('./api.js')
@@ -93,14 +91,18 @@ describe('wrapResult — central 403 translation', () => {
 
         await expect(client.channels.deleteChannel(500)).rejects.toMatchObject({
             code: 'FORBIDDEN',
+            message: 'Twist refused this action: 403 Forbidden.',
+            hints: [
+                'You may not have permission for this action',
+                'Contact your workspace admin, or re-authenticate with `tw auth login` if your token looks wrong',
+            ],
         })
     })
 
     it('prefers INSUFFICIENT_SCOPE over FORBIDDEN when error_string indicates scope', async () => {
         sdkMocks.deleteChannel.mockRejectedValueOnce(
-            Object.assign(new Error('Request failed with status 403'), {
-                httpStatusCode: 403,
-                responseData: { error_string: 'Insufficient scope provided: channels:write' },
+            new TwistRequestError('Request failed with status 403', 403, {
+                error_string: 'Insufficient scope provided: channels:write',
             }),
         )
 
@@ -109,22 +111,18 @@ describe('wrapResult — central 403 translation', () => {
 
         await expect(client.channels.deleteChannel(500)).rejects.toMatchObject({
             code: 'INSUFFICIENT_SCOPE',
+            message: 'This action requires permissions your current token does not have.',
+            hints: ['Run `tw auth login` to re-authenticate with the required scopes'],
         })
     })
 
     it('passes non-403 errors through untranslated', async () => {
-        sdkMocks.deleteChannel.mockRejectedValueOnce(
-            Object.assign(new Error('Request failed with status 500'), {
-                httpStatusCode: 500,
-                responseData: {},
-            }),
-        )
+        const originalError = new TwistRequestError('Request failed with status 500', 500, {})
+        sdkMocks.deleteChannel.mockRejectedValueOnce(originalError)
 
         const { createWrappedTwistClient } = await import('./api.js')
         const client = createWrappedTwistClient('test-token')
 
-        await expect(client.channels.deleteChannel(500)).rejects.toThrow(
-            'Request failed with status 500',
-        )
+        await expect(client.channels.deleteChannel(500)).rejects.toBe(originalError)
     })
 })
--- a/src/lib/api.test.ts
+++ b/src/lib/api.test.ts
@@ -1,19 +1,34 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 
-// Mock the SDK so we can observe how `getWorkspaceUsers` invokes
-// `client.workspaceUsers.getWorkspaceUsers`. The real filtering of
-// removed users lives in the SDK (≥2.8.1), so the contract this test
-// guards is "we pass `includeRemoved` through unchanged."
+// Hoisted mocks — shared across both describe blocks.
 const getWorkspaceUsersMock = vi.hoisted(() => vi.fn().mockResolvedValue([]))
-
-vi.mock('@doist/twist-sdk', () => ({
-    TwistApi: class {
-        workspaceUsers = { getWorkspaceUsers: getWorkspaceUsersMock }
-    },
+const sdkMocks = vi.hoisted(() => ({
+    deleteChannel: vi.fn(),
 }))
+
+vi.mock('@doist/twist-sdk', () => {
+    class TwistApi {
+        channels = { deleteChannel: sdkMocks.deleteChannel }
+        workspaceUsers = { getWorkspaceUsers: getWorkspaceUsersMock }
+        constructor(_token?: string) {}
+    }
+    return {
+        TwistApi,
+        TwistRequestError: class TwistRequestError extends Error {
+            constructor(
+                message: string,
+                public httpStatusCode: number,
+                public responseData?: unknown,
+            ) {
+                super(message)
+            }
+        },
+    }
+})
 
 vi.mock('./auth.js', () => ({
     getApiToken: vi.fn().mockResolvedValue('test-token'),
+    getAuthMetadata: vi.fn().mockResolvedValue({ authMode: 'full' }),
 }))
 
 vi.mock('./permissions.js', () => ({
@@ -30,6 +45,10 @@ vi.mock('./progress.js', () => ({
 }))
 
 import { getWorkspaceUsers } from './api.js'
+
+// ─── getWorkspaceUsers ────────────────────────────────────────────────────────
+// The SDK (≥2.8.1) filters removed users itself; the contract this test guards
+// is "we pass `includeRemoved` through unchanged."
 
 describe('getWorkspaceUsers', () => {
     beforeEach(() => {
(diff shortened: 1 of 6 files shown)


src/lib/api.ts
```diff
@@ -194,6 +194,12 @@ function wrapResult(
                     ['Run `tw auth login` to re-authenticate with the required scopes'],
                 )
             }
+            if (isForbidden(error)) {
```
doistbot, line 200: [P2] This only translates 403s for rejected promise-returning methods that pass through `wrapResult`. `client.batch(...)` doesn’t: it’s a top-level `TwistApi` method, and batched 403s are still surfaced later by `assertBatchData()` / `getOptionalBatchData()` as `API_ERROR` or `BATCH_FAILED`. Since commands like `thread view`, `inbox`, `comment delete`, and `msg delete` rely on batch, the new "central" FORBIDDEN/INSUFFICIENT_SCOPE translation is still incomplete. Either wrap `batch` on the root proxy too, or teach the batch helpers to map 403 responses into the same `CliError`s.

src/lib/api.test.ts
```diff
@@ -0,0 +1,88 @@
...
+
+describe('wrapResult — central 403 translation', () => {
+    beforeEach(() => {
+        vi.clearAllMocks()
+        vi.resetModules()
+    })
+
+    it('translates a plain 403 into a FORBIDDEN CliError', async () => {
+        sdkMocks.deleteChannel.mockRejectedValueOnce(
+            Object.assign(new Error('Request failed with status 403'), {
```
doistbot: [P2] The `vi.mock` block at the top of this file explicitly provides a `TwistRequestError` mock implementation, but these tests ignore it and manually reconstruct the same error shape using `Object.assign`. You can `import { TwistRequestError } from '@doist/twist-sdk'` and use `new TwistRequestError('Request failed with status 403', 403, {})` here and in the tests below. This removes the duplicate error assembly and aligns with how `errors.test.ts` handles SDK errors.

src/lib/errors.ts
```diff
@@ -86,6 +87,21 @@ export function isInsufficientScope(error: unknown): boolean {
...
 }
 
+/**
+ * Check whether an error is any Twist API 403 response.
+ *
+ * Precedence: callers must test `isInsufficientScope` first so OAuth-scope
+ * 403s keep their dedicated `INSUFFICIENT_SCOPE` code and hints; `isForbidden`
+ * is the catch-all fallback for plain workspace-permission 403s.
+ */
+export function isForbidden(error: unknown): boolean {
```
doistbot, line 100: [P2] `isForbidden()` duplicates the same Twist-request shape/status narrowing that `isInsufficientScope()` already does. Please extract the common `httpStatusCode` check into a tiny shared helper (or build one predicate on top of the other) so these classifiers don't drift as more API-status helpers get added.

src/lib/api.ts
```diff
@@ -194,6 +194,12 @@ function wrapResult(
...
+                throw new CliError('FORBIDDEN', 'Twist refused this action: 403 Forbidden.', [
+                    'You may not have permission for this action — workspace admins can perform it',
```
doistbot: [P2] This hint now shows up for every SDK method that goes through `wrapResult`, including read-only calls like `getChannel` or `getWorkspaces`. On those paths, “workspace admins can perform it” is often irrelevant or misleading. Either pass method/mutation context into `wrapResult` and tailor the hint, or keep the centralized 403 copy fully generic.

src/lib/api.test.ts
```diff
@@ -0,0 +1,88 @@
...
+                httpStatusCode: 403,
+                responseData: {},
+            }),
+        )
+
+        const { createWrappedTwistClient } = await import('./api.js')
+        const client = createWrappedTwistClient('test-token')
+
+        await expect(client.channels.deleteChannel(500)).rejects.toMatchObject({
```
doistbot, line 92: [P2] This only pins the `code`. The change here is the centralized friendly translation, so the 403 cases should also assert the user-facing `message` (and ideally `hints`). As written, a regression to the wrong text would still pass. The same gap exists in the insufficient-scope case below.

src/lib/api.test.ts
```diff
@@ -0,0 +1,88 @@
...
+            Object.assign(new Error('Request failed with status 500'), {
+                httpStatusCode: 500,
+                responseData: {},
+            }),
+        )
+
+        const { createWrappedTwistClient } = await import('./api.js')
+        const client = createWrappedTwistClient('test-token')
+
+        await expect(client.channels.deleteChannel(500)).rejects.toThrow(
```
doistbot: [P2] `toThrow('Request failed with status 500')` doesn't actually prove non-403 errors are passed through untouched. A wrapped `CliError` with the same message would still satisfy this. Capture the original error object and assert the rejection is that same instance, or at least assert it's not a `CliError`.

doistbot (commented):
This PR successfully centralizes 403 Forbidden error handling within the API proxy, establishing a solid pattern for unified error translation. The refactoring simplifies command-level logic and sets a great foundation for standardizing other API responses. There are a few areas that need further refinement, specifically extending the translation coverage to batch API methods, adapting the error hints to remain accurate for read-only calls, extracting shared logic in the error predicates and test mocks, and strengthening the test assertions to ensure exact messages and object instances are validated.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F247&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae07ce720-5615-11f1-9565-ec2594c4a0ca%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae07ce720-5615-11f1-9565-ec2594c4a0ca&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/api.ts
lmjabreu: Keeping as-is. Batch responses don't go through the promise rejection path that `wrapResult` intercepts; they come back as `{ code, data }` objects checked by `assertBatchData` / `getOptionalBatchData`. A 403 there would need a separate `response.code === 403` check in those helpers, which is a different fix from what this PR does. Happy to file a follow-up to track it separately.

src/lib/api.test.ts
lmjabreu: Done. Added a static `import { TwistRequestError } from '@doist/twist-sdk'` at the top; since `vi.mock` is hoisted, it resolves to the mock class. All three tests now use `new TwistRequestError(...)` directly instead of `Object.assign`.

src/lib/errors.ts
lmjabreu: Done. Extracted `hasTwistStatusCode(error, status)` as a private helper; both `isForbidden` and `isInsufficientScope` now delegate to it.

src/lib/api.ts
lmjabreu: Done. Dropped the "workspace admins can perform it" line; it only makes sense on write-path 403s. New hints: `'You may not have permission for this action'` / `'Contact your workspace admin, or re-authenticate with `tw auth login` if your token looks wrong'`.

src/lib/api.test.ts
lmjabreu: Done. Both 403 cases now assert `code`, `message`, and `hints`. Updated the hint values to match the copy change from thread 4.

src/lib/api.test.ts
lmjabreu: Done. Captures the original error instance before the mock call and switches to `rejects.toBe(originalError)`. A wrapped `CliError` with the same message string wouldn't satisfy strict object identity.

src/lib/errors.ts
```diff
@@ -63,30 +63,43 @@ export type ErrorCode =
     // Escape hatch for dynamic codes
     | (string & {})
 
+function hasTwistStatusCode(error: unknown, status: number): boolean {
+    return (
+        typeof error === 'object' &&
+        error !== null &&
+        'httpStatusCode' in error &&
+        (error as { httpStatusCode: number }).httpStatusCode === status
```
gnapse:
```suggestion
        typeof error.httpStatusCode === 'number' &&
        error.httpStatusCode === status
```

src/lib/errors.ts
```diff
@@ -63,30 +63,43 @@ export type ErrorCode =
...
  * Works with any error shaped like TwistRequestError (httpStatusCode + responseData).
  */
 export function isInsufficientScope(error: unknown): boolean {
-    if (
-        typeof error !== 'object' ||
-        error === null ||
-        !('httpStatusCode' in error) ||
-        !('responseData' in error)
-    ) {
+    if (!hasTwistStatusCode(error, 403) || !('responseData' in (error as object))) {
```
gnapse: Tell your coding agent to do its best to get rid of the "something as whatever" typecasts. Both this one here and the one in line 83 a few lines below this one. It has to truly validate the shape of the object during runtime instead of typecasting.

src/lib/errors.ts
```diff
@@ -63,30 +63,43 @@ export type ErrorCode =
...
         responseData.error_string.includes('Insufficient scope')
     )
 }
 
+/**
+ * Check whether an error is any Twist API 403 response.
+ *
+ * Precedence: callers must test `isInsufficientScope` first so OAuth-scope
+ * 403s keep their dedicated `INSUFFICIENT_SCOPE` code and hints; `isForbidden`
+ * is the catch-all fallback for plain workspace-permission 403s.
```
gnapse, line 98: What if instead of relying on a comment here we redesign the error types system to avoid this? here's what I suggest:

- create a `getErrorType(error: unknown)` function
- the return type is a restricted string union type with the possible answers
- then `isForbidden` and `isInsufficientScope` can be implemented like this:
    ```js
    function isForbidden(error) {
        return getErrorType(error) === 'forbidden'
    }
    function isInsufficientScope(error) {
        return getErrorType(error) === 'insufficient-scope'
    }
    ```
- Maybe we do not even need these functions, and we just rely on exporting `getErrorType` as the main way to check if an error is of a certain type
- `getErrorType` would have embedded in it the workflow of first checking if it's insufficient scope before returning if it's forbidden.

gnapse (changes requested):

src/lib/errors.ts
lmjabreu: Done. Promoted `hasTwistStatusCode` to a type predicate (`error is { httpStatusCode: number }`). The runtime check now reads `typeof error.httpStatusCode === 'number' && error.httpStatusCode === status` with no cast, and callers get the narrowing for free.

src/lib/errors.ts
lmjabreu: Done. Both casts in `isInsufficientScope` are gone. Because `hasTwistStatusCode` is now a type predicate, TS narrows `error` after the call, so `'responseData' in error` works without `as object`. The `responseData` shape is runtime-validated before reading: `typeof data === 'object' && data !== null && 'error_string' in data && typeof data.error_string === 'string'`. No `as` left in the file.

src/lib/errors.ts
lmjabreu: The fragility you're flagging is real. The precedence rule lives in a doc comment that nothing enforces, so a future caller could test `isForbidden` first and get the wrong code without anything noticing. A `getErrorType` discriminator that bakes the precedence into the implementation would catch that statically.

I'd rather do that as a follow-up PR. This one is scoped to the 403 centralization in `wrapResult`, and the redesign touches the public shape of `errors.ts` (whether `isForbidden`/`isInsufficientScope` stay as wrappers or get deprecated). It would also be the right place to extend the type union for 404/429/5xx, which is on the follow-up list. Doing all the API status helpers in one pass beats sprinkling the change across two PRs.

Does that work, or would you rather block this on the redesign?

gnapse (approved):

src/lib/errors.ts
gnapse: That's ok, addressing it in a follow-up refactoring PR.

PR #248, opened 2026-05-22
test: dedupe shared test setup (console spies, program, channel fixture)

scottlovegrove:
## Summary

Audit + dedupe of repeated test scaffolding across the suite. Three focused commits, each its own concern:

- **`captureConsole()`** (`src/test-helpers/console.ts`) — replaces ~209 inline `vi.spyOn(console, ...).mockImplementation(() => {})` spies across 25 files and removes ~159 manual `mockRestore()` calls (auto-restores via `onTestFinished`).
- **`createTestProgram(register)`** (`src/test-helpers/program.ts`) — collapses 24 identical local `createProgram()` definitions; call sites unchanged via a one-line wrapper. Drops now-unused commander imports.
- **`createChannelFixture()`** (`src/lib/__fixtures__/channels.ts`) — replaces 4 byte-identical `sampleChannel` literals (channel add/remove/set/members), matching the existing accounts-fixture pattern.

Net: **−367 lines** (29 files), no behavior change.

## Deliberately not done

- **`makeTwistClientMock`** — the per-command client mocks are each a different SDK subset; a generic helper would be a no-op `{...overrides}` wrapper (premature abstraction).
- **Shared `apiMocks`/`refsMocks` factories** — real dup exists in the 4 channel-membership files, but `vi.hoisted` runs before imports so a shared factory can't be imported into it (TDZ); the `__mocks__/` alternative would force all 16 api-mocking files onto one global mock surface.
- Single-use entity factories (thread/message/comment/conversation) left in place — relocating them wouldn't reduce duplication.

## Test plan

- [x] `npm run lint:check` clean
- [x] `npm test` — 677 passed (43 files)

### DIFF
--- a/src/commands/channel/add.test.ts
+++ b/src/commands/channel/add.test.ts
@@ -1,5 +1,6 @@
 import { Command } from 'commander'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { captureConsole } from '../../test-helpers/console.js'
 
 const mockGetChannel = vi.fn()
 
@@ -63,7 +64,7 @@ describe('tw channel members add', () => {
             expandedFrom: [],
         })
         const program = createProgram()
-        vi.spyOn(console, 'log').mockImplementation(() => {})
+        captureConsole()
 
         await program.parseAsync([
             'node',
@@ -89,7 +90,7 @@ describe('tw channel members add', () => {
             expandedFrom: [{ groupId: 200, groupName: 'Backend', userIds: [4, 5] }],
         })
         const program = createProgram()
-        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+        const consoleSpy = captureConsole()
 
         await program.parseAsync([
             'node',
@@ -114,7 +115,7 @@ describe('tw channel members add', () => {
             expandedFrom: [],
         })
         const program = createProgram()
-        vi.spyOn(console, 'log').mockImplementation(() => {})
+        captureConsole()
 
         await program.parseAsync([
             'node',
@@ -136,7 +137,7 @@ describe('tw channel members add', () => {
             expandedFrom: [],
         })
         const program = createProgram()
-        vi.spyOn(console, 'log').mockImplementation(() => {})
+        captureConsole()
 
         await program.parseAsync([
             'node',
@@ -158,7 +159,7 @@ describe('tw channel members add', () => {
             expandedFrom: [],
         })
         const program = createProgram()
-        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+        const consoleSpy = captureConsole()
 
         await program.parseAsync([
             'node',
@@ -182,7 +183,7 @@ describe('tw channel members add', () => {
             expandedFrom: [],
         })
         const program = createProgram()
-        const consoleSpy = vi.spyOn(console, 'log').mockImplementation(() => {})
+        const consoleSpy = captureConsole()
 
         await program.parseAsync([
             'node',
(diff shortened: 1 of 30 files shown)


src/test-helpers/console.ts
```diff
@@ -0,0 +1,13 @@
+import { onTestFinished, vi } from 'vitest'
```
doistbot, line 1: [P2] Adding test-only helpers under `src/` pulls them into `npm run build`: `tsconfig.build.json` excludes tests, `__mocks__`, and `__fixtures__`, but not `src/test-helpers`. That means the published `dist/` will now include files like this one that import `vitest`, which is only a devDependency. Please either move `src/test-helpers` out of the build graph or add it to the build excludes.

src/commands/channel/add.test.ts
```diff
@@ -26,25 +28,9 @@ vi.mock('chalk')
...
-    creator: 1,
-    public: true,
-    archived: false,
-    created: new Date(),
-    version: 1,
-    url: 'https://twist.com/a/1/ch/500',
-}
+const createProgram = () => createTestProgram(registerChannelCommand)
+
+const sampleChannel = createChannelFixture()
```
doistbot, line 33: [P2] This shared fixture currently weakens type checking compared to the inline literal it replaces. `createChannelFixture()` takes `Record<string, unknown>` overrides, so misspelled keys and wrong field types in overrides will compile cleanly across all four suites. Please type the helper against the real channel shape (for example `Partial<Channel>` overrides and a `Channel` return type, or a dedicated local fixture type) before reusing it more widely.

doistbot (commented):
This PR effectively cleans up the test suite by deduplicating console spies, program initialization, and channel fixtures into shared helpers. This provides a great boost to maintainability by dropping nearly 400 lines of boilerplate while thoughtfully avoiding premature abstractions. A few minor adjustments are needed to ensure the new test helpers directory is properly excluded from the published build output and to strictly type the channel fixture overrides so we maintain the safety of the original inline literals.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F248&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae9c2ff60-5627-11f1-981d-000797046fc9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae9c2ff60-5627-11f1-981d-000797046fc9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/test-helpers/console.ts
scottlovegrove: Fixed in 9ca584b — added `src/test-helpers` to the `exclude` list in `tsconfig.build.json`. Verified a fresh `npm run build` no longer emits `dist/test-helpers`.

src/commands/channel/add.test.ts
scottlovegrove: Fixed in 9ca584b — `createChannelFixture` now takes `Partial<Channel>` overrides and returns `Channel` (imported from `@doist/twist-sdk`), so misspelled keys and wrong field types in overrides are now caught at compile time. `npm run type-check` passes.

gnapse (approved):

PR #249, opened 2026-05-23
fix(refs): resolve unjoined-but-public channels by name

lmjabreu:
## Summary

- `resolveChannelRef`'s name branch only called `client.channels.getChannels({ workspaceId })`, which is **membership-scoped** and excludes public channels the current user hasn't joined.
- Effect: `tw channel <action> "Public Channel I Haven't Joined"` fails with `CHANNEL_NOT_FOUND`, while `id:N` and URL refs work because `getChannel(id)` isn't membership-filtered.
- Fix merges `getChannels` with `workspaces.getPublicChannels` and dedupes by id. Same shape as `src/commands/channel/list.ts:177-180` already uses for the `--scope public` path.

## Why this exists

The bug was first suspected to be "active vs archived" — that turned out to be wrong: `getChannels({ workspaceId })` already returns both states (Twist's API is tri-state — see `list.ts:163-169`). The real axis is **joined vs unjoined**, which I confirmed live and which also holds for non-admin users (`getPublicChannels` is not admin-gated). Scale check in Doist workspace 1585: 998 public channels, 829 of which I haven't joined — so the bug bites in production, not just edge cases.

## Test plan

- [x] `npm run lint` — clean
- [x] `npm run type-check` — no errors in `refs.ts` / `refs.test.ts` (pre-existing `@doist/cli-core` export errors elsewhere on `main` are unrelated)
- [x] `npx vitest run src/lib/refs.test.ts` — **71/71 pass** (4 new tests: unjoined-public exact match, unjoined-public substring match, dedup of joined-and-public, ambiguity across joined+public)
- [x] **Live smoke** in workspace 1767 (Testing Environment), against a fresh test channel — no existing channels touched:
  1. `tw channel create tw-cli-resolver-smoke-<ts> --workspace 1767` → id 837258
  2. Left the channel via SDK `removeUser`
  3. `tw channel archive "tw-cli-resolver-smoke-<ts>" --workspace 1767` → `{ id: 837258, archived: true }` ✅ (previously: `CHANNEL_NOT_FOUND`)
  4. `tw channel delete id:837258 --workspace 1767 --yes` → cleaned up

Smoke required #246's `channel create/archive/delete` commands to exercise an action on the resolved channel; they were checked out temporarily and reverted after.

## Out of scope

- Help-text cleanup on `tw channel unarchive` in #246's `src/commands/channel/index.ts` — separate trivial follow-up once both this and #246 land.
- `id:` and URL branches — already work for archived and unjoined channels via `getChannel(id)`.

### DIFF
--- a/src/lib/refs.test.ts
+++ b/src/lib/refs.test.ts
@@ -216,6 +216,17 @@ describe('resolveChannelRef', () => {
 
     const mockGetChannel = vi.fn()
     const mockGetChannels = vi.fn()
+    const mockGetPublicChannels = vi.fn()
+
+    /**
+     * For name refs, resolveChannelRef merges joined channels (getChannels — membership-scoped,
+     * includes both active + archived) with public channels (getPublicChannels — workspace-scoped,
+     * finds unjoined-but-public channels). Tests default both to empty unless overridden.
+     */
+    function mockChannelLists(joined: unknown[] = [], publicChannels: unknown[] = []) {
+        mockGetChannels.mockResolvedValue(joined)
+        mockGetPublicChannels.mockResolvedValue(publicChannels)
+    }
 
     beforeEach(() => {
         vi.clearAllMocks()
@@ -223,6 +234,9 @@ describe('resolveChannelRef', () => {
             channels: {
                 getChannel: mockGetChannel,
                 getChannels: mockGetChannels,
+            },
+            workspaces: {
+                getPublicChannels: mockGetPublicChannels,
             },
         })
     })
@@ -265,19 +279,20 @@ describe('resolveChannelRef', () => {
         expect(mockGetChannel).not.toHaveBeenCalled()
     })
 
-    it('resolves exact case-insensitive name match', async () => {
+    it('resolves exact case-insensitive name match against joined channels', async () => {
         const ch = createChannel(10, 'General')
-        mockGetChannels.mockResolvedValue([ch, createChannel(20, 'Leadership')])
+        mockChannelLists([ch, createChannel(20, 'Leadership')])
 
         const result = await resolveChannelRef('general', 1)
 
         expect(mockGetChannels).toHaveBeenCalledWith({ workspaceId: 1 })
+        expect(mockGetPublicChannels).toHaveBeenCalledWith(1)
         expect(result).toEqual(ch)
     })
 
     it('resolves unique substring name match', async () => {
         const ch = createChannel(30, 'Marketing')
-        mockGetChannels.mockResolvedValue([createChannel(10, 'General'), ch])
+        mockChannelLists([createChannel(10, 'General'), ch])
 
         const result = await resolveChannelRef('market', 1)
 
@@ -285,10 +300,7 @@ describe('resolveChannelRef', () => {
     })
 
     it('throws AMBIGUOUS_CHANNEL on multiple substring matches', async () => {
-        mockGetChannels.mockResolvedValue([
-            createChannel(10, 'Engineering'),
-            createChannel(20, 'Engineering-Ops'),
-        ])
+        mockChannelLists([createChannel(10, 'Engineering'), createChannel(20, 'Engineering-Ops')])
 
         await expect(resolveChannelRef('eng', 1)).rejects.toHaveProperty(
             'code',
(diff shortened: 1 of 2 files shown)


scottlovegrove:
@doistbot /review

src/lib/refs.ts
```diff
@@ -230,7 +230,22 @@ export async function resolveChannelRef(ref: string, workspaceId: number): Promi
...
+        // joined (across active + archived). Public channels the user hasn't joined are not
+        // included, so name-resolving e.g. `tw channel archive "Old Public Channel"` would
+        // fail with CHANNEL_NOT_FOUND even though the channel is discoverable. Merge with
+        // getPublicChannels (workspace-scoped, returns all public channels regardless of
+        // membership) and dedupe by id so a joined-and-public channel doesn't match twice.
+        const [joined, publicChannels] = await Promise.all([
+            client.channels.getChannels({ workspaceId }),
+            client.workspaces.getPublicChannels(workspaceId),
+        ])
+        const byId = new Map<number, Channel>()
```
doistbot: [P3] This logic introduces a new `Map`-based approach for deduplication. `src/commands/channel/list.ts` handles this exact same joined/public intersection by using a `Set` of joined IDs. Aligning with that existing pattern reduces drift and simplifies the code:

```typescript
const joinedIds = new Set(joined.map((channel) => channel.id))
const channels = [
    ...joined,
    ...publicChannels.filter((channel) => !joinedIds.has(channel.id))
]
```

src/lib/refs.ts
```diff
@@ -230,7 +230,22 @@ export async function resolveChannelRef(ref: string, workspaceId: number): Promi
...
 
     if (parsed.type === 'name') {
-        const channels = await client.channels.getChannels({ workspaceId })
+        // getChannels is membership-scoped — it returns only channels the current user has
+        // joined (across active + archived). Public channels the user hasn't joined are not
+        // included, so name-resolving e.g. `tw channel archive "Old Public Channel"` would
+        // fail with CHANNEL_NOT_FOUND even though the channel is discoverable. Merge with
+        // getPublicChannels (workspace-scoped, returns all public channels regardless of
+        // membership) and dedupe by id so a joined-and-public channel doesn't match twice.
+        const [joined, publicChannels] = await Promise.all([
```
doistbot, line 239: [P2] This `Promise.all` makes every name-based channel lookup depend on `workspaces.getPublicChannels`, even when the ref is an exact match in the joined list. That widens the failure surface for common lookups and adds a full public-channel fetch to all `resolveChannelRef()` callers. A better fit here is to resolve against `getChannels({ workspaceId })` first and only fetch/merge `getPublicChannels(workspaceId)` when the joined list alone can't produce an exact match.

src/lib/refs.test.ts
```diff
@@ -265,30 +279,28 @@ describe('resolveChannelRef', () => {
...
-    it('resolves exact case-insensitive name match', async () => {
+    it('resolves exact case-insensitive name match against joined channels', async () => {
         const ch = createChannel(10, 'General')
-        mockGetChannels.mockResolvedValue([ch, createChannel(20, 'Leadership')])
+        mockChannelLists([ch, createChannel(20, 'Leadership')])
 
         const result = await resolveChannelRef('general', 1)
 
         expect(mockGetChannels).toHaveBeenCalledWith({ workspaceId: 1 })
+        expect(mockGetPublicChannels).toHaveBeenCalledWith(1)
```
doistbot: [P3] This assertion hard-codes the current `Promise.all` fetch strategy. If `resolveChannelRef()` is later optimized to return after an exact joined-channel match, this test will fail even though the behavior is still correct. The result assertion already covers the contract here; drop the `getPublicChannels` call expectation.

src/lib/refs.test.ts
```diff
@@ -297,13 +309,51 @@ describe('resolveChannelRef', () => {
...
+    it('resolves unjoined-but-public channel by substring', async () => {
+        const publicCh = createChannel(60, 'tw-cli-smoke-test-channel')
+        mockChannelLists([createChannel(10, 'General')], [publicCh])
+
+        const result = await resolveChannelRef('smoke-test', 1)
+
+        expect(result).toEqual(publicCh)
+    })
+
+    it('deduplicates channels appearing in both joined and public lists', async () => {
```
doistbot, line 336: [P2] This test doesn't actually prove deduplication. `matchByName()` returns the first exact case-insensitive match before it checks for ambiguity, so this still passes if the same channel appears twice in the merged list. Use a non-exact query (for example, a substring match) so the test would fail without the dedupe step.

doistbot (commented):
This PR successfully expands name-based channel resolution to include unjoined public channels, fixing a frustrating gap in how the CLI handles larger workspaces. The approach is solid and well-tested, ensuring users can interact with any public channel regardless of their membership status. A few refinements are noted regarding aligning the deduplication logic with existing codebase patterns, deferring the public channel API fetch until necessary to reduce overhead, and tweaking the test suite to better prove deduplication without hardcoding the underlying fetch strategy.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F249&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aed0e0a40-56ac-11f1-8329-1e128f5406f8%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aed0e0a40-56ac-11f1-8329-1e128f5406f8&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/refs.test.ts
lmjabreu: Done. Dropped the `getPublicChannels` and `getChannels` call-shape assertions in 8c1d8f6; the result check is enough.

src/lib/refs.ts
lmjabreu: Keeping as-is. The two-call eager fetch is deliberate: it lets `matchByName` detect cross-list ambiguity when a joined channel and an unjoined public channel share a name. A lazy fallback would silently prefer the joined one on exact-match collision, which I'd rather surface than hide. Open to revisiting if the extra `getPublicChannels` call shows up as a real cost.

src/lib/refs.test.ts
lmjabreu: Done. Switched the dedup test to a substring query (`'eng'`) in 8c1d8f6 so it actually exercises the partial-match counting. The exact `'Engineering'` query would have passed regardless of dedup because `matchByName` returns on the first `.find`.

src/lib/refs.ts
lmjabreu: Done. Switched to the `Set<joinedId>.filter()` pattern from `list.ts:177-185` in 8c1d8f6.

Bloomca (approved):
The code change looks good. I do agree that it can add a bit of noise (in general I assume most of the time we want to look at joined channels), but let's try and see.

lmjabreu:
Thanks @Bloomca. Agreed it's worth watching. If the substring noise shows up on the bigger workspaces, the lazy-fetch fallback doistbot suggested (joined first, widen to public only on a miss) is the natural next step. Merging.

PR #250, opened 2026-05-23
refactor(account): adopt cli-core account command attachers

scottlovegrove:
## Summary
- Bump `@doist/cli-core` 0.20.0 → 0.23.0.
- Replace the hand-rolled `account` commands (`list` / `use` / `remove` / `current`) with cli-core's `attachAccount*Command` (new in 0.22.0); delete `list.ts`, `use.ts`, `remove.ts`, `current.ts`.
- `current` keeps its env/legacy/config output by overriding the store's `activeAccount()` to return `null` for env-token / stuck-legacy sessions, so the attacher routes to its async `onNotAuthenticated` hook (where `isLegacyAuthActive()` can run); a resolved account renders as `source: config`. The v1→v2 migration still fires on `tw account current` via `activeAccount()` → `resolveOverrideSnapshot()`. `active()` and `activeAccount()` are now both env/legacy-aware (removes a latent divergence).
- Preserve twist behavior via attacher hooks: `list` human format (`renderText`), `remove` keyring-fallback warning (`onRemoved`), and the v1→v2 legacy-migration guard (`AUTH_MIGRATION_PENDING`) via a store wrapper since the attachers own the action.
- Forward the new `TokenStore.clear() → ClearedAccount | null` contract through `withUserRefAware` + test mocks.
- Add `CODEBASE.md` repo map (descriptive, matching the cli-core / todoist-cli style).

## ⚠️ Machine-output contract change
Adopting cli-core's conventions changes `account` machine output (`current` is unchanged — same `{…, source}` shapes as before):
| cmd | before | after |
|---|---|---|
| `list --json` | bare array `[{id,label,isDefault}]` | `{accounts:[{account,isDefault}], default}` |
| `use --json` | `{id,label,isDefault}` | `{ok:true, default}` |
| `remove --json` | `{id,label,removed}` | `{ok:true, removed}` |
| `use`/`remove --ndjson` | emitted object | silent (success-action convention) |

This breaks any downstream consumer of the old `account … --json/--ndjson` output. Per maintainer call this ships as a **minor** (account JSON not treated as part of the semver contract), so no `BREAKING CHANGE:` footer.

## Test plan
- [x] `npm run lint:check` clean
- [x] `npm run build` clean
- [x] `npm test` — 677/677 pass (account tests rewritten to the new shapes)
- [x] Smoke-tested built CLI: 4 subcommands registered, `tw account` → list, `list --json` emits the new envelope, `TWIST_API_TOKEN=… account current` → `{source:"env"}`


src/commands/account/index.ts
```diff
@@ -1,42 +1,99 @@
...
-    ).action((ref: string, options: ViewOptions) => useAccount(ref, options, store))
-
-    withJsonFlags(
-        account
-            .command('remove <ref>')
-            .description('Remove a stored account (clears keyring + config entry)'),
-    ).action((ref: string, options: ViewOptions) => removeAccount(ref, options, store))
+    // The list attacher adds `list` without commander's `isDefault`, so wire the
+    // parent default explicitly to keep `tw account` (no subcommand) listing.
+    ;(account as unknown as { _defaultCommandName: string })._defaultCommandName = 'list'
```
doistbot, line 139: [P2] Commander provides a public `defaultCommand()` method for this exact use case (added in v8). Bypassing the type system to mutate the internal `_defaultCommandName` property is brittle and unnecessary. Replace this cast with `account.defaultCommand('list')`.

src/commands/account/index.ts
```diff
@@ -1,42 +1,99 @@
...
+        },
+    })
 }
 
 export function registerAccountCommand(program: Command): void {
     const account = program.command('account').description('Manage stored CLI accounts')
     const store = createTwistTokenStore()
+    const guarded = withLegacyGuard(store)
+
+    attachAccountListCommand(account, {
```
doistbot, line 55: [P1] This block swaps `account list/use/remove` onto cli-core's default emitters, which changes the existing machine-output contract (`list --json` envelope, `use/remove --json` payloads, and `use/remove --ndjson` becoming silent). That will break existing scripts that already consume these commands. Either preserve Twist's current JSON/NDJSON shapes around these attachers, or ship this as an explicit breaking/major change.

src/commands/account/index.ts
```diff
@@ -1,42 +1,99 @@
...
+/**
+ * Guard the v2-store mutators so a pending v1→v2 migration (store empty, legacy
+ * token still live) surfaces AUTH_MIGRATION_PENDING instead of a misleading
+ * empty list / ACCOUNT_NOT_FOUND. cli-core's attachers own the command action,
+ * so the guard lives in the store rather than at a call site. `current` uses the
+ * unguarded store — it reports the legacy/env session rather than erroring.
+ */
+function withLegacyGuard(store: TwistTokenStore): TwistTokenStore {
+    return Object.assign(Object.create(store) as TwistTokenStore, {
+        async list() {
```
doistbot, line 35: [P3] `withLegacyGuard` repeats the same `await assertV2Available(); return store.<method>(...)` wrapper three times. A small local helper to wrap guarded store methods would keep the migration check in one place and avoid drift if another account-store method needs the same gate later.

src/commands/account/index.ts
```diff
@@ -1,42 +1,99 @@
...
-        .option('--json', 'Output as JSON')
-        .option('--ndjson', 'Output as newline-delimited JSON')
+/**
+ * Guard the v2-store mutators so a pending v1→v2 migration (store empty, legacy
+ * token still live) surfaces AUTH_MIGRATION_PENDING instead of a misleading
+ * empty list / ACCOUNT_NOT_FOUND. cli-core's attachers own the command action,
+ * so the guard lives in the store rather than at a call site. `current` uses the
+ * unguarded store — it reports the legacy/env session rather than erroring.
+ */
+function withLegacyGuard(store: TwistTokenStore): TwistTokenStore {
```
doistbot, line 28: [P2] `withLegacyGuard()` moves the legacy-auth check from once-per-command to once-per-store-call. `assertV2Available()` goes through `isLegacyAuthActive()`, which can hit both the secure store and config file; with the cli-core attachers, a single command can now trigger that more than once (for example, `account use --json` needs `setDefault()` and a follow-up `list()` to produce `{ default }`). Please cache the guard promise inside this wrapper, or run the guard once before entering the attacher, so one command doesn't re-read legacy state multiple times.

src/commands/account/account.test.ts
```diff
@@ -197,61 +215,61 @@ describe('account command', () => {
...
 
             await expect(
                 createProgram().parseAsync(['node', 'tw', 'account', 'use', '999']),
             ).rejects.toHaveProperty('code', 'ACCOUNT_NOT_FOUND')
-
-            expect(storeMocks.setDefault).not.toHaveBeenCalled()
         })
 
-        it('matches refs by display name and resolves to the canonical id', async () => {
+        it('resolves a display-name ref to the canonical default id under --json', async () => {
```
doistbot, line 222: [P2] This rewrite locks in the new `--json` envelope for `account use`, but not the other intentional contract change: `account use --ndjson` is now supposed to emit nothing on success. Add a `--ndjson` assertion with no stdout so a future cli-core bump can't silently change that behavior.

doistbot (commented):
This PR effectively adopts the new `cli-core` command attachers to streamline account management while preserving essential twist-specific behaviors. This approach significantly reduces bespoke code and improves the long-term maintainability of the CLI. A few areas for refinement have been noted, particularly around safely addressing the breaking changes to machine output formats, optimizing the legacy authentication guard to prevent redundant state reads, and leveraging Commander's public API for configuration.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F250&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A066ca5d0-56e1-11f1-96b1-c221fc1b94ad%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A066ca5d0-56e1-11f1-96b1-c221fc1b94ad&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/account/index.ts
```diff
@@ -1,42 +1,99 @@
...
 }
 
 export function registerAccountCommand(program: Command): void {
     const account = program.command('account').description('Manage stored CLI accounts')
     const store = createTwistTokenStore()
+    const guarded = withLegacyGuard(store)
+
+    attachAccountListCommand(account, {
+        store: guarded,
+        description: 'List stored CLI accounts',
```
scottlovegrove:
```suggestion
        description: 'List stored Twist accounts',
```

src/commands/account/index.ts
scottlovegrove: commander 14.0.3 has no public `defaultCommand()` method — it doesn't exist in this version (`grep defaultCommand node_modules/commander/lib/command.js` finds only the internal `_defaultCommandName`). The public way to set a default is `.command(name, { isDefault: true })`, which the attacher hardcodes away (`parent.command('list')`). So the cast is the minimal way to keep `tw account` → list. Open to a cleaner approach if one exists in 14.

src/commands/account/index.ts
scottlovegrove: Deliberate — documented in the PR's contract-change table. Maintainer decision: ship as a **minor**; the account `--json`/`--ndjson` output isn't treated as part of the semver contract here. No `BREAKING CHANGE:` footer added.

src/commands/account/index.ts
scottlovegrove: Centralised the gate in a cached `ensureV2()` closure (cd987c2) so the migration check lives in one place. Kept the three thin method overrides rather than a higher-order wrapper to avoid extra indirection.

src/commands/account/index.ts
scottlovegrove: Done in cd987c2 — `ensureV2 = () => (guard ??= assertV2Available())` memoises within the wrapper, so `use --json` (setDefault + follow-up list) reads legacy state at most once per command.

src/commands/account/account.test.ts
scottlovegrove: Added in cd987c2 — `account use --ndjson` now asserts no stdout on success.

src/commands/account/index.ts
scottlovegrove: Applied in ea883ea.

pedroalves0 (commented):
✅

PR #260, opened 2026-05-31
feat(attachments): support uploading files on replies

scottlovegrove:
## What

Adds file-attachment upload to `tw thread reply` and `tw conversation reply`, using the new `attachments` client added in `@doist/twist-sdk` [2.9.0](https://github.com/Doist/twist-sdk-typescript/pull/141).

```bash
tw thread reply 12345 "See attached" --file ./diagram.png
tw conversation reply 12345 "Photos" --file ./a.jpg --file ./b.jpg
tw thread reply 12345 --file ./report.pdf            # file-only, no text
```

## How

- **SDK bump** `@doist/twist-sdk` 2.8.1 → 2.9.0.
- **`src/lib/local-file.ts`** — `openLocalFileAsBlob` resolves a path, verifies readability, and returns a file-backed `Blob` (streamed, never fully buffered), with structured `FILE_NOT_FOUND` / `FILE_READ_ERROR`. Mirrors the helper in `todoist-cli`.
- **`src/lib/attachments.ts`** — `uploadAttachments(files)` validates **every** path before uploading any (a bad path fails fast, no orphaned uploads), then uploads each and returns the `Attachment[]` to splice into the create call.
- `--file` is **repeatable**; a **file-only** post with no text content is allowed.
- `--file` + `--close`/`--reopen` → `CONFLICTING_OPTIONS` (those arg shapes don't accept attachments).

## Auth / scopes (403 handling)

Uploading needs the new `attachments:write` scope. A token from before the scope change gets a 403. Because `attachments.upload` isn't in `KNOWN_SAFE_API_METHODS`, it's treated as mutating and routes through `wrapResult`, so the existing centralized translation kicks in: an "Insufficient scope" 403 becomes `INSUFFICIENT_SCOPE` with the `tw auth login` re-login prompt. Added an `attachments.upload` spinner entry to guarantee the wrap. Scope lists in `auth-provider.ts` now include `attachments:read` / `attachments:write`.

## Not included: `thread create`

Intentionally left out — the SDK's `CreateThreadArgs` has no `attachments` field and the schema is zod `$strip`, so unknown keys are dropped before the request. Supporting it needs an SDK change first.

## Verification

- `type-check`, full suite (**729 tests**, incl. new `local-file` unit tests, `--file` suites for both reply surfaces, and an `attachments.upload` 403 → `INSUFFICIENT_SCOPE` test), oxlint + oxfmt all pass.
- Verified end-to-end against the live API: uploaded a ~780KB JPEG via `tw conversation reply` — message posted with the attachment.

### DIFF
--- a/src/commands/conversation/conversation.test.ts
+++ b/src/commands/conversation/conversation.test.ts
@@ -223,6 +223,30 @@ function createClient({
 
 const createProgram = () => createTestProgram(registerConversationCommand)
 
+// Shared setup for the --file suite: a fresh mock client wired into getTwistClient
+// plus a program. Tests asserting on output call captureConsole() themselves.
+function setupFileTest() {
+    const client = createClient({})
+    apiMocks.getTwistClient.mockResolvedValue(client)
+    return { client, program: createProgram() }
+}
+
+// Registers a temp dir with two files for a --file suite, cleaned up afterwards.
+function useFileFixtures(prefix: string, png: string, pdf: string) {
+    const paths = { dir: '', png: '', pdf: '' }
+    beforeAll(async () => {
+        paths.dir = await mkdtemp(join(tmpdir(), prefix))
+        paths.png = join(paths.dir, png)
+        paths.pdf = join(paths.dir, pdf)
+        await writeFile(paths.png, 'png-bytes')
+        await writeFile(paths.pdf, 'pdf-bytes')
+    })
+    afterAll(async () => {
+        await rm(paths.dir, { recursive: true, force: true })
+    })
+    return paths
+}
+
 describe('conversation implicit view', () => {
     beforeEach(() => {
         vi.clearAllMocks()
@@ -795,32 +819,16 @@ describe('conversation done', () => {
 })
 
 describe('conversation reply --file', () => {
-    let tmpDir: string
-    let filePng: string
-    let filePdf: string
-
-    beforeAll(async () => {
-        tmpDir = await mkdtemp(join(tmpdir(), 'tw-convo-reply-'))
-        filePng = join(tmpDir, 'photo.png')
-        filePdf = join(tmpDir, 'doc.pdf')
-        await writeFile(filePng, 'png-bytes')
-        await writeFile(filePdf, 'pdf-bytes')
-    })
-
-    afterAll(async () => {
-        await rm(tmpDir, { recursive: true, force: true })
-    })
+    const files = useFileFixtures('tw-convo-reply-', 'photo.png', 'doc.pdf')
 
     beforeEach(() => {
         vi.clearAllMocks()
     })
 
     it('uploads the file and attaches it to the message', async () => {
-        const client = createClient({})
-        apiMocks.getTwistClient.mockResolvedValue(client)
+        const { client, program } = setupFileTest()
         const consoleSpy = captureConsole()
 
-        const program = createProgram()
         await program.parseAsync([
             'node',
             'tw',
@@ -829,7 +837,7 @@ describe('conversation reply --file', () => {
             '42',
             'See attached',
             '--file',
-            filePng,
+            files.png,
         ])
 
         expect(client.attachments.upload).toHaveBeenCalledTimes(1)
@@ -847,10 +855,8 @@ describe('conversation reply --file', () => {
     })
 
     it('attaches multiple repeated --file values', async () => {
-        const client = createClient({})
-        apiMocks.getTwistClient.mockResolvedValue(client)
+        const { client, program } = setupFileTest()
 
-        const program = createProgram()
         await program.parseAsync([
             'node',
             'tw',
@@ -859,9 +865,9 @@ describe('conversation reply --file', () => {
             '42',
             'two files',
             '--file',
-            filePng,
+            files.png,
             '--file',
-            filePdf,
+            files.pdf,
         ])
 
         expect(client.attachments.upload).toHaveBeenCalledTimes(2)
@@ -872,11 +878,9 @@ describe('conversation reply --file', () => {
     })
 
     it('allows a file-only message with no text content', async () => {
-        const client = createClient({})
-        apiMocks.getTwistClient.mockResolvedValue(client)
+        const { client, program } = setupFileTest()
 
-        const program = createProgram()
-        await program.parseAsync(['node', 'tw', 'conversation', 'reply', '42', '--file', filePng])
+        await program.parseAsync(['node', 'tw', 'conversation', 'reply', '42', '--file', files.png])
 
         expect(client.conversationMessages.createMessage).toHaveBeenCalledWith(
             expect.objectContaining({ content: '', attachments: expect.any(Array) }),
@@ -886,10 +890,8 @@ describe('conversation reply --file', () => {
     })
 
     it('errors with FILE_NOT_FOUND for a missing path and does not send', async () => {
-        const client = createClient({})
-        apiMocks.getTwistClient.mockResolvedValue(client)
+        const { client, program } = setupFileTest()
 
-        const program = createProgram()
         await expect(
             program.parseAsync([
                 'node',
@@ -899,7 +901,7 @@ describe('conversation reply --file', () => {
                 '42',
                 'x',
                 '--file',
-                join(tmpDir, 'missing.png'),
+                join(files.dir, 'missing.png'),
             ]),
         ).rejects.toMatchObject({ code: 'FILE_NOT_FOUND' })
 
@@ -908,11 +910,9 @@ describe('conversation reply --file', () => {
     })
 
     it('does not upload during --dry-run but lists the attachment', async () => {
-        const client = createClient({})
-        apiMocks.getTwistClient.mockResolvedValue(client)
+        const { client, program } = setupFileTest()
         const consoleSpy = captureConsole()
 
-        const program = createProgram()
         await program.parseAsync([
             'node',
             'tw',
@@ -921,12 +921,12 @@ describe('conversation reply --file', () => {
             '42',
             'preview',
             '--file',
-            filePng,
+            files.png,
             '--dry-run',
         ])
 
         expect(client.attachments.upload).not.toHaveBeenCalled()
         expect(client.conversationMessages.createMessage).not.toHaveBeenCalled()
-        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining(filePng))
+        expect(consoleSpy).toHaveBeenCalledWith(expect.stringContaining(files.png))
     })
 })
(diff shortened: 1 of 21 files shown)


src/commands/thread/index.ts
```diff
@@ -55,6 +56,7 @@ Examples:
         )
         .option('--close', 'Close the thread after replying')
         .option('--reopen', 'Reopen the thread after replying')
+        .option('--file <path>', 'Attach a file (repeatable; content optional)', collect, [])
```
doistbot, line 59: <big><kbd>🟠 P1</kbd></big>  Per `AGENTS.md`, whenever options are added in `src/commands/`, the `SKILL_CONTENT` in `src/lib/skills/content.ts` must be updated to match. Please add the new `--file` option to the command references for both `thread reply` and `conversation reply` in `content.ts` so that AI agents can discover it.

src/commands/conversation/reply.ts
```diff
@@ -12,34 +14,43 @@ export async function replyToConversation(
...
         printDryRun('send message to conversation', {
             Conversation: String(conversationId),
-            Content: preview,
+            Attach: hasFiles ? files.join(', ') : undefined,
+            Content: preview || undefined,
         })
         return
     }
 
+    const attachments = hasFiles ? await uploadAttachments(files) : undefined
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This uploads files before the command has validated that the target conversation exists or is writable. `tw conversation reply <bad-or-forbidden-id> --file ./a.png` will upload the attachment successfully and then fail in `createMessage`, leaving an unattached upload behind. Please preflight the conversation (for example with `client.conversations.getConversation(conversationId)`) before calling `uploadAttachments`.

src/lib/auth-provider.ts
```diff
@@ -44,6 +44,7 @@ export const READ_WRITE_SCOPES = [
     'groups:remove',
     'search:read',
     'notifications:read',
+    `attachments:write`,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `READ_WRITE_SCOPES` should include `attachments:read` too. As written, a normal `tw auth login` grants upload permission but not the corresponding read scope, so read-write auth is no longer a superset of read-only auth for attachments.

src/lib/attachments.ts
```diff
@@ -0,0 +1,26 @@
...
+ * upload starts, so a bad path fails fast without leaving a partial set of
+ * uploaded-but-unreferenced attachments behind.
+ */
+export async function uploadAttachments(files: string[]): Promise<Attachment[]> {
+    // Validate every path first so a bad one fails before we upload anything.
+    const opened = await Promise.all(files.map((file) => openLocalFileAsBlob({ file })))
+
+    const client = await getTwistClient()
+    const attachments: Attachment[] = []
+    for (const { blob, fileName } of opened) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Missed concurrency: these files are uploaded sequentially. Consider using `Promise.all` to upload them concurrently, which will reduce latency when multiple files are attached while still preserving the array order for the message payload.

```ts
return Promise.all(
    opened.map(({ blob, fileName }) => client.attachments.upload({ file: blob, fileName }))
)
```

src/lib/attachments.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Upload one or more local files and return the created {@link Attachment}s,
+ * ready to splice into the `attachments` array of `comments.createComment`,
+ * `threads.createThread`, or `conversationMessages.createMessage`.
+ *
+ * All paths are validated (existence + readability) up front, before any
+ * upload starts, so a bad path fails fast without leaving a partial set of
+ * uploaded-but-unreferenced attachments behind.
+ */
+export async function uploadAttachments(files: string[]): Promise<Attachment[]> {
+    // Validate every path first so a bad one fails before we upload anything.
```
doistbot, line 16: <big><kbd>🟡 P2</kbd></big>  The key guarantee in this helper—"validate every path before uploading any"—isn't covered by the new tests. The command suites only exercise an all-missing path, so a regression that uploads the first file and then fails on a later missing file would still pass while leaving orphaned uploads behind. Add a mixed valid+missing case (or a small `uploadAttachments()` unit test) that asserts `attachments.upload` is never called when any path is invalid.

doistbot (commented):
Thanks scottlovegrove for your contribution 😄 🙌. The attachment implementation cleanly integrates file validation and the new SDK endpoints.

Few things worth tightening:
- Update `SKILL_CONTENT` in `src/lib/skills/content.ts` with the new `--file` option so AI agents can discover it.
- Preflight the conversation in `conversation reply` before uploading attachments to prevent orphaned files on invalid or forbidden targets.
- Switch to `Promise.all` for concurrent multi-file uploads, and add a test covering the mixed valid/missing path case to enforce the "no partial uploads" guarantee.
- Add `attachments:read` to `READ_WRITE_SCOPES` so read-write auth remains a proper superset of read-only auth.

I also included a few optional follow-up notes in the details below.

<details>
<summary>Optional follow-up notes (2)</summary>

- **[P3] src/lib/attachments.ts:8**: The shared helper’s contract now says its return value is ready for `threads.createThread`, but this PR intentionally does not support thread-create attachments because that SDK call drops the field. Please remove `threads.createThread` from the docstring so the lib API doesn’t advertise a shape callers can’t actually use.
- **[P3] src/commands/thread/thread.test.ts:1414**: This proves a file-only reply succeeds, but it doesn't actually guard the new "skip the editor for file-only replies" behavior. Because `openEditor` is mocked to resolve `''`, the test still passes if a regression opens the editor before posting. Please assert `openEditor` is not called here; the same gap exists in the conversation suite.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F260&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A587b12d0-5cdb-11f1-9676-945262ad10c4%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A587b12d0-5cdb-11f1-9676-945262ad10c4&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/commands/thread/index.ts
scottlovegrove: Done — added the `--file` option to both `thread reply` and `conversation reply` in `src/lib/skills/content.ts`; `check:skill-sync` passes. (16b8481)

src/commands/conversation/reply.ts
scottlovegrove: Done — `conversation reply` now preflights with `client.conversations.getConversation(conversationId)` before uploading whenever files are attached, so an invalid/forbidden target fails before any upload. (16b8481)

src/lib/auth-provider.ts
scottlovegrove: Done — added `attachments:read` to `READ_WRITE_SCOPES` (and normalized the scope quoting so it's no longer a template literal). (16b8481)

src/lib/attachments.ts
scottlovegrove: Done — switched to `Promise.all` for concurrent uploads; the returned array still preserves input order for the message payload. (16b8481)

src/lib/attachments.ts
scottlovegrove: Done — added `src/lib/attachments.test.ts` with a mixed valid+missing case asserting `attachments.upload` is never called when any path is invalid (plus an order-preservation test). (16b8481)

src/commands/conversation/conversation.test.ts
```diff
@@ -766,3 +793,140 @@ describe('conversation done', () => {
...
+        await program.parseAsync([
+            'node',
+            'tw',
+            'conversation',
+            'reply',
+            '42',
+            'See attached',
+            '--file',
+            filePng,
+        ])
```
scottlovegrove, line 841: Nearly all of these new tests use the same setup, can we reuse?

src/commands/thread/thread.test.ts
```diff
@@ -1319,3 +1337,290 @@ describe('thread done', () => {
...
+        await program.parseAsync([
+            'node',
+            'tw',
+            'thread',
+            'reply',
+            '500',
+            'See attached',
+            '--file',
+            filePng,
+        ])
```
scottlovegrove, line 1386: Same with these test setups

src/lib/local-file.ts
```diff
@@ -0,0 +1,51 @@
...
+import { open } from 'node:fs/promises'
+import { basename, resolve } from 'node:path'
+import { CliError } from './errors.js'
+
+export interface LocalFileOptions {
+    /** Path to the file on disk (relative paths resolve against cwd). */
+    file: string
+    /** Optional override for the upload's user-facing filename. Defaults to `basename(file)`. */
+    fileName?: string
+}
```
scottlovegrove: Prefer types over interfaces

src/commands/conversation/conversation.test.ts
scottlovegrove: Done — extracted a `setupFileTest()` helper (fresh mock client + `getTwistClient` wiring + program) and a `useFileFixtures()` helper for the temp-dir lifecycle, so each test is now just `const { client, program } = setupFileTest()`. (91f6083)

src/commands/thread/thread.test.ts
scottlovegrove: Done — same `setupFileTest()` + `useFileFixtures()` helpers applied to both the `thread reply --file` and `thread create --file` suites. (91f6083)

src/lib/local-file.ts
scottlovegrove: Done — `LocalFileOptions` is now a `type` alias instead of an `interface`. (91f6083)

PR #266, opened 2026-06-05
fix(channels): Include unjoined public channels in public-channel allowlist

### LINKED ISSUE #263: getPublicChannelIds is membership-scoped — same bug #249 fixed in refs.ts
## Summary

`getPublicChannelIds` in [`src/lib/public-channels.ts`](https://github.com/Doist/twist-cli/blob/main/src/lib/public-channels.ts#L7-L19) builds its public-channel allowlist from `client.channels.getChannels({ workspaceId })` **only**. That endpoint is **membership-scoped** — it excludes public channels the current user hasn't joined.

This is the same bug class that #249 fixed in `resolveChannelRef` (`src/lib/refs.ts`), but in a file #249 didn't touch.

```ts
// src/lib/public-channels.ts:12
const channels = await client.channels.getChannels({ workspaceId })
const publicIds = new Set<number>()
for (const ch of channels) {
    if (ch.public) publicIds.add(ch.id)
}
```

## Impact

`getPublicChannelIds` feeds two callers:

1. **`assertChannelIsPublic`** (`public-channels.ts:25`) — gates thread access. A thread in a **public channel the user hasn't joined** is wrongly rejected with `NOT_FOUND: "This thread belongs to a private channel."`, even though it's public and readable. This is the exact `threads/get` workaround path that the newsletter-crawler skill relies on.

2. **`filterVisibleSearchResults`** (`src/lib/search-helpers.ts:183`) — client-side filter on search results. Practically masked today because the server `/api/v3/search` index is itself membership-scoped (returns zero for unjoined channels), so there's nothing to wrongly filter out. But it's latent: if/when search scope widens server-side, this filter would silently drop unjoined-public results.

Scale check (from #249): Doist workspace 1585 has 998 public channels, 829 of them unjoined for the test user — so this bites in production, not just edge cases.

## Suggested fix

Mirror #249: merge `getChannels` with `workspaces.getPublicChannels` and dedupe by id. Same shape already used in `src/commands/channel/list.ts:177-180` for the `--scope public` path.

## Out of scope

- Server-side search index scope (`tw search` only indexing joined channels) — that's a Twist API concern, not fixable in the CLI. Context: https://github.com/Doist/doist-os/pull/245#issuecomment-4606838736

engfragui:
## Overview

`getPublicChannelIds` built its allowlist from `getChannels` alone, which is membership-scoped and excludes public channels the user hasn't joined. This wrongly rejected threads in unjoined-but-public channels as private via assertChannelIsPublic, and left a latent masking bug in filterVisibleSearchResults.

Merge `getChannels` with `workspaces.getPublicChannels` and dedupe by id, mirroring the #249 fix in resolveChannelRef.

## Reference

- Fixes #263
- Similar to https://github.com/Doist/twist-cli/pull/249

## Test Plan

- `npm run build`
- Via the UI, visit an unjoined channel (i.e. I picked "z.Cooking" but you do you)
- Copy the link of any thread inside that unjoined channel
- Extract the threadId from the URL (it's the last id in the URL)
- `node dist/index.js thread view id:<threadId>`
- [ ] Verify that the thread renders normally (instead of `NOT_FOUND: This thread belongs to a private channel`)

### DIFF
--- a/src/lib/public-channels.ts
+++ b/src/lib/public-channels.ts
@@ -9,11 +9,6 @@ export async function getPublicChannelIds(workspaceId: number): Promise<Set<numb
     if (cached) return cached
 
     const client = await getTwistClient()
-    // getPublicChannels is workspace-scoped: it returns every public channel (active and
-    // archived, joined and unjoined). getChannels is membership-scoped, so building the
-    // allowlist from it alone wrongly excluded public channels the user hasn't joined — a
-    // thread in such a channel was rejected as private (#263). getPublicChannels is a complete
-    // superset of the public channels getChannels would surface, so it's the only call needed.
     const publicChannels = await client.workspaces.getPublicChannels(workspaceId)
     const publicIds = new Set(publicChannels.map((ch) => ch.id))
     publicChannelCache.set(workspaceId, publicIds)
--- a/src/lib/public-channels.ts
+++ b/src/lib/public-channels.ts
@@ -9,10 +9,21 @@ export async function getPublicChannelIds(workspaceId: number): Promise<Set<numb
     if (cached) return cached
 
     const client = await getTwistClient()
-    const channels = await client.channels.getChannels({ workspaceId })
+    // getChannels is membership-scoped — it returns only channels the current user has joined,
+    // so public channels the user hasn't joined are excluded. Merge with getPublicChannels
+    // (workspace-scoped, returns all public channels regardless of membership) so a thread in
+    // an unjoined-but-public channel isn't wrongly rejected as private. Mirrors resolveChannelRef
+    // in refs.ts.
+    const [joined, publicChannels] = await Promise.all([
+        client.channels.getChannels({ workspaceId }),
+        client.workspaces.getPublicChannels(workspaceId),
+    ])
     const publicIds = new Set<number>()
-    for (const ch of channels) {
+    for (const ch of joined) {
         if (ch.public) publicIds.add(ch.id)
+    }
+    for (const ch of publicChannels) {
+        publicIds.add(ch.id)
     }
     publicChannelCache.set(workspaceId, publicIds)
     return publicIds
--- a/src/lib/public-channels.ts
+++ b/src/lib/public-channels.ts
@@ -9,22 +9,13 @@ export async function getPublicChannelIds(workspaceId: number): Promise<Set<numb
     if (cached) return cached
 
     const client = await getTwistClient()
-    // getChannels is membership-scoped — it returns only channels the current user has joined,
-    // so public channels the user hasn't joined are excluded. Merge with getPublicChannels
-    // (workspace-scoped, returns all public channels regardless of membership) so a thread in
-    // an unjoined-but-public channel isn't wrongly rejected as private. Mirrors resolveChannelRef
-    // in refs.ts.
-    const [joined, publicChannels] = await Promise.all([
-        client.channels.getChannels({ workspaceId }),
-        client.workspaces.getPublicChannels(workspaceId),
-    ])
-    const publicIds = new Set<number>()
-    for (const ch of joined) {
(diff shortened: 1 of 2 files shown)


engfragui:
@doistbot /review

doistbot (commented):
Thanks for putting together this fix for unjoined public channels, Francesca 😎 👊.

Few things worth tightening:
- The `getChannels` API call is likely redundant and can be removed, as `getPublicChannels` already returns all public channels regardless of membership (unless you specifically need to capture archived joined channels).

I also included a few optional follow-up notes in the details below.

<details>
<summary>Optional follow-up note (1)</summary>

- **[P3] src/lib/public-channels.ts:17**: This is now the third place that reimplements the "joined channels + `getPublicChannels` + dedupe by id" rule (alongside `resolveChannelRef` and `listChannels`). Since this PR exists because those call sites drifted apart, it would be better to extract a shared helper for this workspace-channel visibility logic and build the ID set from that, so future fixes only need to land once.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Ftwist-cli&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Ftwist-cli%2Fpull%2F266&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A9d4e7020-60d5-11f1-9d47-989403e39a15%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A9d4e7020-60d5-11f1-9d47-989403e39a15&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/lib/public-channels.ts
```diff
@@ -9,11 +9,22 @@ export async function getPublicChannelIds(workspaceId: number): Promise<Set<numb
     if (cached) return cached
 
     const client = await getTwistClient()
-    const channels = await client.channels.getChannels({ workspaceId })
+    // getChannels is membership-scoped — it returns only channels the current user has joined,
+    // so public channels the user hasn't joined are excluded. Merge with getPublicChannels
+    // (workspace-scoped, returns all public channels regardless of membership) so a thread in
+    // an unjoined-but-public channel isn't wrongly rejected as private. Mirrors resolveChannelRef
+    // in refs.ts.
+    const [joined, publicChannels] = await Promise.all([
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Since `getPublicChannels` returns all public channels regardless of membership, fetching `getChannels` here is likely an unnecessary API call. 

`resolveChannelRef` needs `getChannels` because it must also resolve joined *private* channels. However, since this function only collects public channel IDs, you can likely drop `getChannels` (and the `Promise.all`) entirely and just populate the Set directly from `publicChannels`.

*(Note: If `getPublicChannels` excludes archived channels and you intentionally need `getChannels` to capture archived joined public channels, this approach is fine—but otherwise the extra call can be safely removed).*

scottlovegrove (approved):

